PR #4, opened 2017-06-17
Map Infrastructure  and Activity Submission Marker Refinement

SethGreylyn:
This pull request comprises my initial work setting up the WellsMap object and putting an interactive map onto the 'Submit Activity' page. The 'Submit Activity' map and attendant functionality is not finished (there are unresolved questions about how to populate the lat/long fields, how to store them, and how to update them, among others), but the patterns are there to build upon.

### DIFF
--- a/gwells/views.py
+++ b/gwells/views.py
@@ -20,8 +20,6 @@ class HelloWorldView(generic.ListView):
         """
         return WellYieldUnit.objects.order_by('-sort_order')
 
-
-
 def well_search(request):
     well_results = None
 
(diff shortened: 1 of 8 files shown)


gwells/views.py
```diff
@@ -20,8 +20,6 @@ def get_queryset(self):
         """
         return WellYieldUnit.objects.order_by('-sort_order')
 
-
```
SethGreylyn: This whitespace deletion was accidental, and can be safely reverted.

PR #10, opened 2017-07-01
Well Entry Client Warnings

SethGreylyn:
This work comprises the DOD of https://trello.com/c/MY2W2JQ9/194-spatial-map-warning-error-message. Note that 'validation' is a misnomer in a technical sense, since invalid fields flag a warning to be shown as a yellow banner atop the coordinate region of the form, but this warning does not (yet) hinder form submission. Proper validation (in the technical sense of preventing form submission) should be considered once the data model is more settled.

NOTE ABOUT CUSTOM-FORM.CSS: It appears my IDE altered each line of the file in an invisible way (e.g., by changing the line endings) when I modified it to add the class, and I didn't catch this on commit, so Git thinks the whole file has been (substantively) changed. I suggest editing the file before accepting the pull request by copying my only real change (which is the .lat-long-error-pane class), overwriting custom-form.css with its old contents, and re-adding my class.


gwells/static/gwells/css/custom-form.css
```diff
@@ -1,604 +1,611 @@
-/** It's all about forms **/
```
SethGreylyn: This file should probably be manually reverted before accepting the PR.

PR #37, opened 2017-07-20
Search Identify Rectangle Improvements

SethGreylyn:
This work comprises feedback on the 'identify' rectangle's behaviour on the search map. Now the actual rectangle drawn persists.

The map still zooms and centres all the actual wells returned from the search, though, which may be jarring (though I personally prefer it). Let me know if you think it would be better to leave the map fit to the rectangle instead of to the wells within the rectangle.


gwells/forms.py
```diff
@@ -88,7 +88,7 @@ def helper(self):
                 # start_lat_long and end_lat_long are programatically generated
                 # based on an identifyWells operation on the client.
                 Hidden('start_lat_long', ''),
-                Hidden('end_lat_long', '')
+                Hidden('end_lat_long', ''),
```
SethGreylyn: This change can be reverted.

PR #63, opened 2017-09-13
Comment out Google Analytics

rstens:
Analytics does not need to be activated right now and it has actually cause the dev instance to become visible through Google search.


garywong-bc:
After resolving conflict manually (Google <script> section commented out).

PR #85, opened 2017-09-29
increase wait repeat time to 1 second

rstens:


rstens:
Wrong update, ignore this one.

PR #178, opened 2017-11-16
Ticket#338

fieranmason:
changes to address QA issues in screen details and some in well development


fieranmason:
made another change in response to PO feedback

PR #270, opened 2017-12-08
Patch to casing internal diameter change

fieranmason:


garywong-bc:
Did you do a git pull upstream master and run db-replicate.sh locally?  On my ENV, now it gives me:

Tuples only is off.
...importing wells_casings data
psql:migrate_casings.sql:38: ERROR:  column "diameter" of relation "gwells_casing" does not exist
LINE 7: diameter

You didn’t answer my question from before.. is this a ‘rename back to a previous’ name?
gary

fieranmason:
not a rename back no --- forward only and on PO request

fieranmason:
have you run the new migration locally

error was that diameter doesn't exist in the model -- that sounds like a
missing migration.

PR #276, opened 2017-12-10
To workaround DB crash when 'out of space'... 

garywong-bc:
Refactored and optimized all SQL, but unable to fully test as Openshift Pipeline was down.  If this breaks any functional test, then we can quickly revert back to prior commit:
a55b7bde71f269c98553b4bdf16cae04fa1a23d8


garywong-bc:
@fieranmason  hi.. plse do a git pull upstream master to ensure you have the changes, and then test out the db-replicate.sh in your local env.  thanks

PR #344, opened 2018-01-09
tweak ReadMe for starting URL.

spatialbits:
Fixes url in Step 8 in ReadMe.


rstens:
No, this is not valid as we have an environment variable installed in our development environments which is set to "gwells". So the instruction is correct given that context.

garywong-bc:
N/A

spatialbits:
Missed that. Right-o.

PR #374, opened 2018-01-23
Implemented Keycloak auth prototype. Installed registry place holder app.

fieranmason:


garywong-bc:
@fieranmason   After another review with PO and UX and DA, there are major changes to the model.py.  Can you resubmit a PR with JUST the auth test?  It shouldn't depend on any registry/model.py content as we'll be redoing it.  It CAN depend on /registry app URLS or views though.  thx

PR #517, opened 2018-03-12
Patch table hide

fieranmason:


fieranmason:
[EPIC#WellSummary | Add no information available message instead of tables with no data |1| !](https://trello.com/c/onCzo9y2/737-epicwellsummary-add-no-information-available-message-instead-of-tables-with-no-data-1)

PR #558, opened 2018-03-28
Hot Patch - break out db_replicate() into two distinct steps

garywong-bc:
To avoid 'Past EOF' database error.


garywong-bc:
@DerekRoberts  Can you make a note of this hotpatch commit?  We'll need to backport it into the Developer branch.  thanks.

PR #564, opened 2018-03-29
Feature/better error map server

Sybrand:
- Added error message for missing map tiles, and moved error message to top of page to be consistent with other error messages.
- Fix some python/pep8 linting issues.
- Added to .gitignore for direnv and moved vscode ignore to correct place.
- Fixed bootstrap formatting bug on existing error messages.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -112,6 +112,7 @@ gwells.pyproj*
 /bin/
 *.sublime-project
 *.sublime-workspace
+.vscode/settings.json
 
 #functional-tests
 functional-tests/.gradle/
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -128,8 +129,9 @@ x.sh
 
 #virtual env
 /deactivate
-.vscode/settings.json
```
DerekRoberts: Good catch!  That's definitely an IDE file and not virtual environment.

Sybrand:
[EPIC#Search | Better Error when Map Server is down |2| !!!](https://trello.com/c/rOx4iuU2/851-epicsearch-better-error-when-map-server-is-down-2)

PR #565, opened 2018-03-29
fixes for Registries search page (clean up styles and text)

stephenhillier:
* Made a copy of the gwells/base.html template for registries, so that unnecessary css and js files can be removed (for example, leaflet.css and esri.js are not needed for the search page)
* removed custom-form.css from Registries search page. This means text on the search form is now slightly larger, consistent with gov.bc.ca style guidelines
* Results table is no longer displayed by default (it is hidden until a search is made)


frontend/src/registry/App.vue
```diff
@@ -17,10 +17,195 @@ export default {
...
 <style>
-/* @import url('https://unpkg.com/bootstrap@3.3.7/dist/css/bootstrap.min.css');
 @import url('https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css');
 @import '../../../gwells/static/gwells/css/ie10-viewport-bug-workaround.css';
 @import '../../../gwells/static/gwells/css/main.css';
-@import '../../../gwells/static/gwells/css/custom-form.css';
-@import '../../../gwells/static/gwells/css/custom-main.css'; */
+@import '../../../gwells/static/gwells/css/custom-main.css';
+
+.row.no-pad {
```
Sybrand: Shouldn't we be using an import here?

frontend/src/registry/App.vue
```diff
@@ -17,10 +17,195 @@ export default {
...
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+    background-image: none;
+    border: 1px solid;
+    border-radius: 0px;
+}
+.btn-primary {
+    color: #fff;
+    background-color: #38598a;
```
Sybrand: Can we look at using something like Sass? For example for defining variables, e.g. then we can change colours in a single location.

registries/templates/registries/base.html
```diff
@@ -0,0 +1,254 @@
...
+    <!-- STANDARD CSS -->
+    <link type="text/css" href="{% static 'gwells/css/main.css' %}" rel="stylesheet" />
+
+    <!-- HTML5 shim and Respond.js for IE8 support of HTML5 elements and media queries -->
+    <!--[if lt IE 9]>
+      <script src="https://oss.maxcdn.com/html5shiv/3.7.2/html5shiv.min.js"></script>
+      <script src="https://oss.maxcdn.com/respond/1.4.2/respond.min.js"></script>
+    <![endif]-->
+
+    <!-- <link rel="stylesheet" href="https://ajax.googleapis.com/ajax/libs/jqueryui/1.12.1/themes/base/jquery-ui.css" type="text/css" media="all" /> -->
```
Sybrand: Can we get rid of this dead code?

PR #583, opened 2018-04-06
Updating API tests for new model relationships

stephenhillier:


stephenhillier:
Updated with db_column names for person (on Register model), organization (on Person model)

PR #589, opened 2018-04-09
Environment variables - non build breaking

Sybrand:
Implementing our learning of last week, here is a version that gets environmental variables without breaking the build by default. By default REQUIRE_ENV_VARIABLES is set to False, and use of default values / not setting variables will result in a warning not an exception. Switch REQUIRE_ENV_VARIABLES to True, and warnings turn into exceptions.

- Now possible to enter "strict" mode by setting REQUIRE_ENV_VARIABLES=True
- Some PEP8 compliancy changes.
- Move of settings.py to settings/__init__.py (in order to break out settings.base module).

@DerekRoberts @gwongBC : this version shouldn't break the build, but please indicate to me a good time to merge after you've approved. Once you've gotten rid of all warnings, we can talk about setting REQUIRE_ENV_VARIABLES=True by default.


Sybrand:
[EPIC#Other | OS Environment/OpenShift Secrets |2| !!!](https://trello.com/c/ZWApME1s/872-epicother-os-environment-openshift-secrets-2)

PR #590, opened 2018-04-10
Registries API - better registrations & applications endpoints

stephenhillier:
* improved queries generated by Django (fewer database trips when loading data) (views.py)
* adjusted fields/columns included in serialized data (serializers.py)
* added delete requests to API tests to clean up the applications and registrations created during testing
* fixed indentation in createtestuser management command


registries/models.py
```diff
@@ -409,7 +409,7 @@ class RegistriesApplicationStatus(AuditModel):
     application = models.ForeignKey(
         RegistriesApplication,
         db_column='application_guid',
-        on_delete=models.PROTECT,
+        on_delete=models.CASCADE,
```
Sybrand: @garywong-bc this cascade opens up a can of worms. We need to talk about marking records as deleted instead of deleting them, then the protect vs. cascade issue goes away.

PR #591, opened 2018-04-10
Additions to Registries Data Loading

garywong-bc:


garywong-bc:
Ignoring as we need to do a new 'makemigrations', making the existing \registries\migrations\000x_xx files obsolete.

PR #592, opened 2018-04-11
Reset migrations and updated FT counts on Registries.

garywong-bc:


registries/models.py
```diff
@@ -154,7 +154,7 @@ class Meta:
 
     def __str__(self):
         return '%s %s' % (self.first_name, self.surname)
-    
+
     @property
     def name(self):
         return '%s %s' % (self.first_name, self.surname)
```
stephenhillier: return str(self) ?

PR #596, opened 2018-04-12
Updated FT record counts.

garywong-bc:


garywong-bc:
Not waiting for code review as purely FT record counts.

PR #598, opened 2018-04-13
Missing get_env_var() and updated FT test results.

garywong-bc:


gwells/minioClient.py
```diff
@@ -1,5 +1,6 @@
 import os
 from minio import Minio
+from gwells.settings.base import get_env_variable
```
Sybrand: @garywong-bc my bad - sorry.

PR #622, opened 2018-04-19
Registries search page fixes

stephenhillier:
Fixes to Registries search based on UX feedback.
* when making a search, browser scrolls to search results
* search results reset when search completed
* better sorting of name and company (including sort by surname by default)
* more tests for registries search table and other components


frontend/src/registry/components/search/SearchTable.vue
```diff
@@ -14,8 +14,8 @@
           <tr v-if="drillers.results && drillers.results.length"
               v-for="(driller, index) in drillers.results"
               :key="`tr ${driller.person_guid} ${index}`" :id="`registry-table-row-${index}`">
-            <td>
-              <div><b>{{ driller.first_name }} {{ driller.surname }}</b></div>
+            <td :id="`drillerName${index}`">
+              <div><b>{{ driller.surname }}, {{ driller.first_name }}</b></div>
```
Sybrand: I think we should use a class here instead of using a bold tag.

stephenhillier:
Now using bootstrap class "font-weight-bold"

PR #629, opened 2018-04-23
Person add and Organization add pages/views (frontend and backend)

stephenhillier:
#### Registries frontend:
* New screens for adding a person record and adding an organization record

#### Registries backend:
* for Person and Organization endpoints, HTTP DELETE requests now set expired_date to current date instead of deleting the record (logical delete). Records with non-null expired_date are filtered out of queries
* Person endpoint supports nested Register and ContactInfo objects (automatically created and linked to person if included in request body)


frontend/src/registry/components/people/PersonDetail.vue
```diff
@@ -245,15 +262,13 @@ export default {
 
             if (app.status_set && app.status_set.length) {
               statusPriority.forEach((code) => {
-                console.log('checking status', code)
                 const statusLevel = app.status_set.findIndex((item) => {
                   return item.status === code
                 })
 
                 if (~statusLevel) {
```
Sybrand: Is there a good reason to be using a tilda here?

frontend/src/registry/components/search/SearchHome.vue
```diff
@@ -260,12 +276,14 @@ export default {
       }
     },
     sortTable (sortCode) {
-      if (this.searchParams.ordering && this.searchParams.ordering.length && this.searchParams.ordering[0] !== '-') {
-        this.searchParams['ordering'] = `-${sortCode}`
+      if (
```
Sybrand: Can this be on one line?

registries/views.py
```diff
@@ -199,6 +198,18 @@ def get_serializer_class(self):
...
 
+    def destroy(self, request, *args, **kwargs):
+        """
+        Set expired_date to current date
+        """
+        try:
+            instance = self.get_object()
+            instance.expired_date = timezone.now()
+            instance.save()
+        except Http404:
```
Sybrand: Why would this 404? Should we not at least be logging an error here?

PR #645, opened 2018-04-26
Correct loading of companies attached to Register.

garywong-bc:


garywong-bc:
Only SQL and CSV changes, so I'm not asking for a dev review.

PR #649, opened 2018-04-26
Fixes to Registries replication.

garywong-bc:


garywong-bc:
Just SQL and CSV changes...

PR #661, opened 2018-04-30
Removed padding on well detail page (and fixed double bootstrap load)

Sybrand:
- Fixed padding issue on well detail
- Removed loading of bootstrap from django template, now relying on header to load it.

### DIFF
--- a/gwells/static/gwells/css/main.css
+++ b/gwells/static/gwells/css/main.css
@@ -18,6 +18,16 @@ a {
     line-height: 1.42857143;
 }
 
+#main-content,
+.col-sm-12,
+.col-md-4,
+.col-md-6,
+.col-md-7,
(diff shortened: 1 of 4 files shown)


Sybrand:
[EPIC#WellSummary | Text alignment |2| !!](https://trello.com/c/Fd09of1Z/956-epicwellsummary-text-alignment-2)

gwells/static/gwells/css/main.css
```diff
@@ -18,6 +18,16 @@ a {
     line-height: 1.42857143;
 }
 
+#main-content,
```
stephenhillier: Can we first try .no-gutters on the rows or add p-0 to individual columns? Wondering if this will affect other parts of the site

PR #668, opened 2018-05-07
Restructured Database folders  

garywong-bc:
Restructured Database folders to have /gwells, /wellsearch in addition to already existing /registries.


garywong-bc:
Database changes only... so not waiting for reviews.

PR #677, opened 2018-05-10
consolidated database/cron/scripts/*.sh to a central folder

garywong-bc:
... under openshift/scripts/


garywong-bc:
Not waiting for review as it's a simple file move...

PR #689, opened 2018-05-17
Classifications and qualifications

Sybrand:


Sybrand:
[EPIC#Registry | Add new applicant - Classification and Qualifications |5| !!! [gherkin]](https://trello.com/c/5xbTW0Fz/795-epicregistry-add-new-applicant-classification-and-qualifications-5-gherkin)

app/frontend/src/registry/components/people/ApplicationAdd.vue
```diff
@@ -1,218 +1,156 @@
...
+      let result = {
+        issuer: [{value: null, text: 'Please select an option'}],
+        classifications: [],
+        qualifications: []
       }
-      for (let i = 0; i < qualList.length; i++) {
-        if (qualList[i] in this.editClassification.qualCodes) {
-          this.editClassification.qualCodes[qualList[i]] = true
-        }
+      if (this.$options.propsData.activity in this.$store.getters.drillerOptions) {
```
stephenhillier: Since we are already using mapGetters (in the computed properties) you could use `...mapGetters(['loading', 'drillerOptions'])` and then reference this.drillerOptions

app/frontend/src/registry/components/people/ApplicationAdd.vue
```diff
@@ -1,218 +1,156 @@
...
+        result.qualifications = options.WellClassCode.map((item) => { return {'text': item.description, 'value': item.registries_well_class_code} })
       }
+      return result
+    },
+    subactivityMap () {
+      return this.$options.propsData.activity in this.$store.getters.drillerOptions ? this.$store.getters.drillerOptions[this.$options.propsData.activity].SubactivityCode : []
     }
+  },
+  created () {
+    this.fetchDrillerOptions({activity: this.$options.propsData.activity})
```
stephenhillier: This component already has some declared properties `props: ['value', 'activity']` so you can access them with `this.activity` and `this.value`

app/frontend/src/registry/components/people/ApplicationAdd.vue
```diff
@@ -1,218 +1,156 @@
...
-          </div>
-        </form>
-      </div>
+          </b-col>
+        </b-row>
+      </p>
+      <p v-else class="card-text">
+        <b-form-group label="Certification">
+          <b-row>
+            <b-col md="2">Issued by</b-col>
```
stephenhillier: We should include the `label-for="the-form-field-id"` attribute on form-groups so bootstrap-vue can generate the aria/screenreader attributes. I also think this works best with one form input per form group but I could be mistaken- try running it through Lighthouse

app/frontend/src/registry/components/people/ApplicationAdd.vue
```diff
@@ -1,218 +1,156 @@
...
+            <b-col md="8">
+              <b-form-checkbox-group class="fixed-width" :options="formOptions.qualifications" v-model="qualificationForm.qualifications" disabled>
+              </b-form-checkbox-group>
+            </b-col>
+          </b-row>
+        </b-form-group>
+        <b-form-group label="Date application received">
+          <b-row>
+            <b-col>
+              <datepicker format="yyyy-MM-dd" v-model="qualificationForm.status_set[0].effective_date" required></datepicker>
```
Sybrand: I don't know why I'm not just using <b-form-input v-model="qualificationForm.status_set[0].effective_date" type="date" required/>, I'll make that change.

app/frontend/src/registry/components/people/ApplicationAdd.vue
```diff
@@ -1,218 +1,156 @@
...
-            <input type="text" id="removalDate" placeholder="Select removal date" v-model="editClassification.application.removalDate">
-          </div>
-        </div>
-      </form>
-    </fieldset>
   </div>
 </template>
 
 <script>
+import Datepicker from 'vuejs-datepicker'
```
Sybrand: I don't think we need to use this - can just use the std. html date picker: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

api-tests/local_newman.sh
```diff
@@ -5,13 +5,42 @@
...
+#     export GWELLS_API_TEST_CLIENT_SECRET"""
 # - Create Django test user (requires env variables from previous step):
 #     python manage.py createtestuser
 # - Install newman:
 #     npm install -g newman
 # - Run script:
 #     ./local_newman.sh
 
+
+if [ -z ${GWELLS_API_TEST_USER+x} ]; then
```
DerekRoberts: If this is going to be updated regularly it might be better to iterate though a list.  Otherwise this is looking thorough!

E.g. 
```
THINGS=(
	"GWELLS_API_TEST_USER"
	"GWELLS_API_TEST_PASSWORD"
	"GWELLS_API_BASE_URL"
	"GWELLS_API_TEST_AUTH_SERVER"
	"GWELLS_API_TEST_CLIENT_SECRET"
)
for t in ${THINGS[@]}
do
	[  ! -z $t ] || echo "Unset: $t" && exit
done

File: app/frontend/src/registry/components/people/ApplicationAdd.vue (hunk 5 above)
Comment (stephenhillier) [reply app/frontend/src/registry/components/people/ApplicationAdd.vue]: per MDN, the html datepicker is not supported by IE or Safari, and unfortunately some users might need IE11. We should try out our options in IE11

File: app/frontend/src/registry/components/people/ApplicationAdd.vue (hunk 3 above)
Comment (Sybrand) [reply app/frontend/src/registry/components/people/ApplicationAdd.vue]: Form-group documentation is confusing, it does say it's "purpose is to pair form controls with a legend or label". But it doesn't say how would have label-for apply to multiple form controls. I've changed the HTML to just have one label per input.

File: app/frontend/src/registry/components/people/ApplicationAdd.vue (hunk 5 above)
Comment (Sybrand) [reply app/frontend/src/registry/components/people/ApplicationAdd.vue]: Sticking with the datepicker then.

File: api-tests/local_newman.sh (hunk 6 above)
Comment (Sybrand) [reply api-tests/local_newman.sh]: It's less code, so I like it - I'm not quite the bash expert though, so I've done this:

```ENV_VARS=(
    "GWELLS_API_TEST_USER"
    "GWELLS_API_TEST_PASSWORD"
    "GWELLS_API_BASE_URL"
    "GWELLS_API_TEST_AUTH_SERVER"
    "GWELLS_API_TEST_CLIENT_ID"
    "GWELLS_API_TEST_CLIENT_SECRET"
)

for env_var in ${ENV_VARS[@]}
do
    if [ -z ${!env_var+x} ]; then 
        echo "$env_var is unset"
        exit
    fi
done
```

PR #700, opened 2018-05-23
Added commit status per stage, defaults to squash commit for CI

cvarjao:


Jenkinsfile
```diff
@@ -25,19 +25,28 @@ private static String stackTraceAsString(Throwable t) {
     return sw.toString()
 }
 
-def _stage(String name, Map context, Closure body) {
+String stageStatusContext(String stageName){
```
cvarjao: This utility function should probably be moved somewhere else

PR #710, opened 2018-05-26
Convert CSV to JSON

garywong-bc:


app/registries/fixtures/proofofagecode.json
```diff
@@ -0,0 +1,44 @@
+[
```
Sybrand: @garywong-bc isn't this duplication? these codes are already in codetables.json

garywong-bc: yes, but I realize that you are using this in your PR, and that TEST and PROD will need to be be manually hit with a 'python manage.py loaddata proofofagecode.json' (there are no complete resets in TEST/PROD DB's).

In a future release, I'll be dropping the proofofage.json and all the CSV'.. part of my 'expand and contract' methodology for data migrations.

PR #714, opened 2018-05-29
Feature/pipeline/mid hook command

cvarjao:


openshift/backend.dc.json
```diff
@@ -149,12 +149,55 @@
...
+                                    },
+                                    {
+                                        "name": "PGHOST",
+                                        "value": "gwells-pgsql${NAME_SUFFIX}"
+                                    }
+                                ]
+                            }
+                        },
+                        "mid": {
+                            "failurePolicy": "Retry",
```
cvarjao: @garywong-bc , is `post-deply.sh` idempotent? meaning, can it safely be re-run if it fails?

gwongBC: yes it can be re-run.   That's why I switched to Stephen's Django management method to create admin accounts and groups; it will not fail if they already exist.

PR #715, opened 2018-05-30
[1005] Pipeline retry on functional tests

DerekRoberts:


DerekRoberts:
@cvarjao Done!  waitUntil is now wrapping try-catch.

Jenkinsfile
```diff
@@ -415,12 +415,28 @@ for(String envKeyName: context.env.keySet() as String[]){
...
+                                    }else{
+                                        sh './gradlew -DchromeHeadlessTest.single=WellDetails chromeHeadlessTest'
+                                    }
+                                    isDone=true
+                                } catch (ex) {
+                                    echo "${stackTraceAsString(ex)}"
+                                    echo "DEV - Functional Tests Failed - Wait one minute and retry once"
+                                    sleep 60
+                                }
+                                if ( attempts >= attemptsMax ){
```
cvarjao: I think we may need to move this if block inside the catch block above, and re-throw the exception, otherwise it will just fail silently.
``if ( attempts >= attemptsMax ){
throw ex
}
``

DerekRoberts: Done!  Good catch.

Jenkinsfile
```diff
@@ -422,19 +422,21 @@ for(String envKeyName: context.env.keySet() as String[]){
...
-                                    sleep 60
-                                }
-                                if ( attempts >= attemptsMax ){
-                                    isDone=true
+                                    if ( attempts < attemptsMax ){
+                                        echo "DEV - Functional Tests Failed - Wait one minute and retry once"
+                                        sleep 60
+                                    } else {
+                                        echo "DEV - Functional Tests Failed - Retry Failed"
+                                        isDone=true
```
cvarjao: change `isDone=true` to `throw ex`

PR #742, opened 2018-06-13
[1037] Fix for Python unit tests and false positive pipeline status at test stage

stephenhillier:
* disabled parallel testing for Python unit tests
* added `set -e` to script that kicks off tests (exit with error on failing tests)


stephenhillier:
[EPIC#Pipeline | False Passes on Unit Tests |1| !!!](https://trello.com/c/KZuvcc7e/1037-epicpipeline-false-passes-on-unit-tests-1)

PR #756, opened 2018-06-18
[1065] Remove redundant fields and tables from model

Sybrand:
- Fields removed from registries_register: status, register_removal_reason, register_removal_date, registration_date
- Tables removed: registries_application_status, registries_status_code
- Redundant post deploy steps removed
- Tests updated to account for removed fields
- View & serializers updated to account for removed fields


Sybrand:
[EPIC#Registry | Remove redundant fields and tables |1| !!!](https://trello.com/c/4WeNtOx1/1065-epicregistry-remove-redundant-fields-and-tables-1)

app/registries/views.py
```diff
@@ -292,12 +293,22 @@ def get_queryset(self):
             else:
```
stephenhillier: Even though it won't make a practical difference (due to the "if user not staff" check on line 299), I think it would still be better to avoid the code on 291-292 from running if user not staff

stephenhillier:
Closing due to deployment issue, will reopen under new PR.

PR #768, opened 2018-06-21
Bug/1000 exempted well driller

garywong-bc:
Will also require a data fix on TEST and PROD:
```
delete from registries_subactivity_code where registries_subactivity_code = 'WELL';
```


garywong-bc:
oops.. wrong branch.

PR #773, opened 2018-06-25
[1000] Minor changes to classification detail/edit

stephenhillier:
* removed references to drilling (like "qualified to drill") from Well Pump Installer classifications
* (minor bug) clear loading state if form submission failed (for example, on a bad request due to too many characters in an input box)

### DIFF
--- a/app/frontend/src/registry/components/people/ApplicationAddEdit.vue
+++ b/app/frontend/src/registry/components/people/ApplicationAddEdit.vue
@@ -54,7 +54,7 @@
           </b-row>
           <b-row>
             <b-col md="8">
-              <b-form-group label="Qualified to drill" class="font-weight-bold">
+              <b-form-group :label="`Qualified${activity === 'DRILL' ? ' to drill' : '' }`" class="font-weight-bold">
                 <b-form-checkbox-group class="fixed-width font-weight-normal" :options="formOptions.qualifications" v-model="qualificationForm.qualifications" disabled>
                 </b-form-checkbox-group>
               </b-form-group>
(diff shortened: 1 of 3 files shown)


stephenhillier:
[EPIC#Registry | Issues for Fixing |2| !!!](https://trello.com/c/F5sWNsTV/1000-epicregistry-issues-for-fixing-2)

app/frontend/src/registry/components/people/PersonDetail.vue
```diff
@@ -580,6 +580,8 @@ export default {
         this.closeApplication(key)
         this.savingApplication = false
         this.updateRecord()
+      }).catch(() => {
+        this.savingApplication = false
```
Sybrand: We need to discuss a logging framework. Ignoring exceptions is bad. We should be doing something along the lines of:
`logger.error(e)`
Ensuring that errors like these end up with our log aggregator, end get logged to consoles when were in development mode.

Could we for now at least do a console.error(e) so that we don't just ignore it?

app/frontend/src/registry/components/people/ApplicationAddEdit.vue
```diff
@@ -54,7 +54,7 @@
           </b-row>
           <b-row>
             <b-col md="8">
-              <b-form-group label="Qualified to drill" class="font-weight-bold">
+              <b-form-group :label="`Qualified${activity === 'DRILL' ? ' to drill' : '' }`" class="font-weight-bold">
```
Sybrand: Do you have a nice guideline of when to do this re. using a computed value?

PR #799, opened 2018-07-05
[1063]  Document storage

Sybrand:
- Refactoring lots of code out into "wells" from "gwells" app (working in progress).
- Vue component to display additional documents (private and public).
- Fix to keycloak login to stop page reloads.
- Document api (minio) and tests added.
- Additional permissions for viewing private documents.


Sybrand:
[EPIC#WellSummary | Add links to internal-only documents |5| !!! [business value ranking:25 ]](https://trello.com/c/IzDBNuCY/768-epicwellsummary-add-links-to-internal-only-documents-5-business-value-ranking25)

app/frontend/src/wells/components/Documents.vue
```diff
@@ -0,0 +1,101 @@
...
+        // The major downside of this is that there's going to be a "popup-blocked" message somewhere.
+        const a = document.createElement('a')
+        a.target = '_blank'
+        a.download = file.name
+        a.href = response.data.url
+        document.body.appendChild(a)
+        a.click()
+      })
+    }
+  }
```
stephenhillier: Code no longer used

app/frontend/src/common/authenticate.js
```diff
@@ -0,0 +1,76 @@
...
+          token,
+          refreshToken,
+          idToken }
+        ).success((result) => {
+          store.commit(SET_KEYCLOAK, instance)
+          if (instance.authenticated) {
+            // We may have been authenticated, but the token could be expired.
+            if (instance.isTokenExpired(0)) {
+              // The token is expired!
+              result = false // As far as we're concerned, this means authentication failed.
```
stephenhillier: possibly reject promise instead?

app/frontend/src/wells/components/Documents.vue
Sybrand: Removed

app/frontend/src/common/authenticate.js
Sybrand: I'm leaving as is right now - mirroring keycloak having true/false for the init - but it's something to relook. We could probably just get rid of this promise entirely.

PR #800, opened 2018-07-06
Update Python Dockerfile, node and npm

DerekRoberts:


docker-images/gwells-python-test/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openshift/python:3.5
+FROM openshift/python:3.6
```
stephenhillier: do we still need this file?

DerekRoberts: I don't think so.  Let's do some cleanup before wrapping this task.

PR #805, opened 2018-07-10
Reordered dependency for set unlogged

DerekRoberts:


DerekRoberts:
Single commit rolled into PR 806.  No need for both.

PR #807, opened 2018-07-10
[1000] Replaced cert_no of grand-parent with N/A

garywong-bc:
So that the test fixtures line up with the new cert_no ('N/A' instead of 'grand-parent').


garywong-bc:
obsolete

PR #819, opened 2018-07-16
[1173] Simple view for construction submission reports

stephenhillier:
Backend:
  * refactor submissions API to be in it's own Django app folder (submissions)
  * template for loading submissions web application
  * API tests: simple requests for basic submissions functionality

Frontend:
  * scaffolded new "submissions" app in frontend folder.
  * bundles to be built for prod are now defined in `build/webpack.prod.conf.js` instead of `webpack.base.conf.js` (this allows us to avoid building all bundles for development)
  * webpack development server now builds only **submissions** app (can be changed in `build/webpack.dev.conf.js`)
  * some refactoring to reduce the dependency on the registry application for logging in.  note: the standalone header still imports the registry store for the Keycloak functions. A future task could create an auth-only store module that all apps can import into their own store.


app/frontend/build/webpack.dev.conf.js
```diff
@@ -93,7 +97,13 @@ const devWebpackConfig = merge(baseWebpackConfig, {
     // in a separate chunk, similar to the vendor chunk
     // see: https://webpack.js.org/plugins/commons-chunk-plugin/#extra-async-commons-chunk
     new webpack.optimize.CommonsChunkPlugin({
-      names: ['header', 'registry', 'documents', 'footer'],
+      names: [
+        'header',
+        'registry',
+        'documents',
+        'footer',
+        'submissions'
```
Sybrand: The order that these are specified in is actually important. Please move submissions above footer.

app/frontend/build/webpack.base.conf.js
```diff
@@ -21,12 +21,6 @@ const createLintingRule = () => ({
 
 module.exports = {
   context: path.resolve(__dirname, '../'),
-  entry: {
```
Sybrand: I think we need this in all three our environments, so I'd rather not re-produce it in dev, test and prod.

app/frontend/src/common/authenticate.js
```diff
@@ -46,7 +50,8 @@ export default {
           refreshToken,
           idToken }
         ).success((result) => {
-          store.commit(SET_KEYCLOAK, instance)
```
Sybrand: Is there a reason we can't just use the imported values?

app/frontend/src/common/common.scss
```diff
@@ -0,0 +1,66 @@
...
+  font-size: 14px;
+  background-color: #f1f1f1
+}
+body {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  background-color: #f1f1f1
+}
+body.registries-body {
```
Sybrand: Is registries really common? Should this not reside in the registries app?

app/frontend/src/submissions/store/index.js
```diff
@@ -0,0 +1,46 @@
...
+    [SET_KEYCLOAK] (state, payload) {
+      state.keycloak = payload
+    }
+  },
+  actions: {},
+  getters: {
+    userRoles (state) {
+      if (state.keycloak && state.keycloak.authenticated) {
+        return {
+          view: (state.keycloak.hasRealmRole('gwells_admin') || state.keycloak.hasRealmRole('wells_viewer')),
```
Sybrand: This looks like common code, why would we put it under submissions?

app/submissions/templates/submissions/base.html
```diff
@@ -0,0 +1,88 @@
+{% load static %}
```
Sybrand: I'm assuming again this is just a shortcut to get things going? There's a lot of code here that can be shared between registries and submissions.

app/frontend/build/webpack.base.conf.js
stephenhillier: If we don't define which app to build for the dev environment, it will try to serve all of them. Only the prod build needs to include all apps.

app/frontend/src/common/authenticate.js
stephenhillier: The submissions app will have it's own store so this file can't import the registry's store.
This and another comment could be resolved by creating a store module for auth-related stuff (keycloak instance and roles) that all apps can include- I'll look into that.

We could also talk about storing the keycloak instance on the Vue instance and access it with `this.$keycloak`

app/frontend/build/webpack.base.conf.js
Sybrand: It needs the entry-points defined though though, or else webpack won't build the bundles that python uses. Like what you've done in webpack.dev.conf.js for submissions.

app/frontend/src/submissions/router/index.js
```diff
@@ -0,0 +1,21 @@
+import Vue from 'vue'
```
Sybrand: Needs license

PR #822, opened 2018-07-19
Container environment for local development

stephenhillier:
* run with `docker-compose up`
* includes postgres database, app server at localhost:8000 and frontend dev server at localhost:8080
* installs small subset of fixtures


docker-images/gwells-develop/webapp/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM node:8-alpine
```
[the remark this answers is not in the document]
stephenhillier: I understand but we aren't always able to match minor/patch releases. We can get a node 8.11 image from docker hub (node:8-alpine is currently 8.11) but I did not find 8.9 (or it's not advertised as available anymore).

docker-compose.yml
```diff
@@ -0,0 +1,58 @@
...
+      APP_CONTEXT_ROOT: gwells
+      DJANGO_SECRET_KEY: secret
+      SESSION_COOKIE_SECURE: "True"
+      CSRF_COOKIE_SECURE: "True"
+      ENABLE_DATA_ENTRY: "True"
+      ENABLE_GOOGLE_ANALYTICS: "False"
+      ENABLE_ADDITIONAL_DOCUMENTS: "False"
+      POSTGRES_SERVICE_PORT: 5432
+      DJANGO_ADMIN_URL: admin
+    command: /bin/bash -c "sleep 3 &&
```
DerekRoberts: This section might be easier to manage from within the Dockerfile.

stephenhillier: I don't remember the exact issue with that but I think it had to do with being dependent on a database connection being available.

PR #825, opened 2018-07-24
Feat/1173 constr

Sybrand:
Backend:

- refactor submissions API to be in it's own Django app folder (submissions)
template for loading submissions web application
- API tests: simple requests for basic submissions functionality

Frontend:

- scaffolded new "submissions" app in frontend folder.
- bundles to be built for prod are now defined in build/webpack.prod.conf.js instead of webpack.base.conf.js (this allows us to avoid building all bundles for development)
- webpack development server now builds only submissions app (can be changed in build/webpack.dev.conf.js)
- some refactoring to reduce the dependency on the registry application for logging in. note: the standalone header still imports the registry store for the Keycloak functions. A future task could create an auth-only store module that all apps can import into their own store.


Sybrand:
[EPIC#Edit | Views for each report |5| !!!](https://trello.com/c/yKKkZnmS/1173-epicedit-views-for-each-report-5)

PR #830, opened 2018-07-26
Release 1.36.0 - might need bug/1180_minio_url_clean before merging to master!

Sybrand:


Sybrand:
@garywong-bc sorry, only saw your comment re. db permissions after merging. I'll look into that and make required changes somewhere in sprint 37.

PR #833, opened 2018-07-26
[931] Update python, node, npm and base image

DerekRoberts:


app/.s2i/bin/assemble
```diff
@@ -51,6 +51,10 @@ elif [[ -f setup.py ]]; then
 fi
 
 cd frontend
+npm cache clean -f
+npm install n
+source ~/.bashrc
+n 8.11.3
```
Sybrand: What us n ?

DerekRoberts: It's a node version manager.  Opinion?
https://www.npmjs.com/package/n

garywong-bc:
“N” seems fine but I use https://github.com/creationix/nvm/blob/master/README.md

PR #840, opened 2018-07-30
[1146] Add edit button on well summary page

Sybrand:
Created a "summary" Vue component under wells - hopefully this can one day become the home of the entire summary.


app/frontend/src/wells/summary.js
```diff
@@ -0,0 +1,37 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+import 'babel-polyfill'
+import Vue from 'vue'
+import Vuex from 'vuex'
+import { store } from '@/registry/store'
```
stephenhillier: we can use the src/common/auth store now

app/frontend/src/wells/summary.js
```diff
@@ -0,0 +1,37 @@
...
+new Vue({
+  el: '#summary',
+  store,
+  components: {
+    Summary
+  },
+  template: '<Summary/>'
+})
+
+Vue.config.devtools = true
```
stephenhillier: we should disable this in production (I think deleting this line will do the trick, if not `NODE_ENV !== 'production'`)

Sybrand: Same deal for header, footer & documents - have changed there also.

app/frontend/src/wells/summary.js
Sybrand: I've changed this for the document storage also

PR #849, opened 2018-08-01
[1178] Refactor: move django to app/backend

Sybrand:


Sybrand:
[EPIC#Other | Refactor project directories |3| !!!](https://trello.com/c/WFwZ3mn1/1178-epicother-refactor-project-directories-3)

app/backend/gwells/settings/__init__.py
```diff
@@ -266,7 +266,7 @@
     'DEFAULT': {
         'CACHE': not DEBUG,
         'BUNDLE_DIR_NAME': 'bundles/',  # must end with slash
-        'STATS_FILE': os.path.join(BASE_DIR, 'frontend/webpack-stats.json')
+        'STATS_FILE': os.path.join(BASE_DIR, '../frontend/webpack-stats.json')
```
stephenhillier: i wonder if we can put this file into the backend dir?

Sybrand: Yes, but I think we should do that as part of cleaning up the django/vue build dependancy.

PR #855, opened 2018-08-05
[doc] update with new directory structure

cedar-technologies:
Hi, I have updated the setup with the change to put the django app in backend folder. I have put the path in front of the cmd to be executed.
Let me know if this makes sense.

Thanks
Alex

### DIFF
--- a/setup/README.md
+++ b/setup/README.md
@@ -56,8 +56,7 @@ Please use either these steps or the install script.
 
     Change directory to repo, then add remote (upstream) origin.
 
-    ```
-    cd gwells
+    ```
     git remote add upstream https://github.com/bcgov/gwells.git
     ```
 
@@ -125,7 +124,7 @@ Please use either these steps or the install script.
     Install Python requirements (file in root of repo).
 
     ```
-    pip3 install -U -r ../requirements.txt
+    app/backend pip3 install -U -r requirements.txt
     ```
 
 5.  ##### Config variables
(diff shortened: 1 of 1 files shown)


stephenhillier:
Hi Alex,
Thanks for another submission. This file needed an update.

Can we keep these preformatted code lines executable (in other words, instead of adding "app/frontend" to each line, put `cd app/frontend` at the top of the code block)?

```
cd app/backend
python3 manage.py loaddata gwells-codetables
python3 manage.py loaddata wellsearch-codetables registries-codetables
python3 manage.py loaddata wellsearch registries
```

cedar-technologies:
Hi @stephenhillier,
yes we can put it at the top of the block it avoids repetition ;-), also what's the diff between python / python3 (sorry new to python) for windows it's in the path so I don't need to specify that.

Sybrand:
@cedar-technologies ```python3``` is pretty much always going to point to python 3. ```python``` may or may not point to python 3 depending on how your system is set up. 

e.g. on my mac by default ```python``` points to python 2:
```
which python
/usr/local/bin/python

which python3
/Library/Frameworks/Python.framework/Versions/3.6/bin/python3
```

VS.

on my mac in a virtualenv, ```python``` points to python 3:
```
which python
/Users/sstrauss/Workspace/gwells/app/backend/.direnv/python-3.6.6/bin/python
```
The argument here is that it's better to be explicit to avoid confusion.

PR #875, opened 2018-08-14
[1179] Pipeline Optimizations and Allocations

DerekRoberts:


Jenkinsfile
```diff
@@ -424,8 +456,6 @@ for(String envKeyName: context.env.keySet() as String[]){
                 openshift.withProject(projectName){
                     podName=openshift.selector('pod', ['deploymentconfig':deploymentConfigName]).objects()[0].metadata.name
                 }
-                // Run migrate
-                sh "oc exec '${podName}' -n '${projectName}' -- bash -c 'cd /opt/app-root/src/backend && pwd && python manage.py migrate'"
```
Sybrand: Remind me, how do we know for sure that migrate has already been run?

DerekRoberts: We don't.  That's going back in.  Good catch.

PR #876, opened 2018-08-14
[650] epicwellsummary jump links 3

cedar-technologies:
Hi,

I have added the quick links on the wells details page. I did my best for it to look ok and to add as minimum as I could. it was not specified anything for mobile so I left it resized as the other columns...

Couple of suggestion:
* lots of repetition in the well-details template (I've added more 😄 ), could be refactored to create small reusable elements...
* the main.css file is huge, I have put my changes at the bottom, there is also lot of repetition that makes it hard to follow.
* the css should be applied by classes, it is difficult when it changes the rendering of base elements. e.g. bootstrap behavior is changed...

Let me know what you think ;-)
Alex

### DIFF
--- a/app/backend/gwells/static/gwells/css/main.css
+++ b/app/backend/gwells/static/gwells/css/main.css
@@ -1686,8 +1686,6 @@ img.back-to-top:hover {
     .contentPageMainColumn .no-print {
         display: none;
     }
-    
-
 }
 
 .alert-success:before {
@@ -2338,4 +2336,3 @@ h5.twitterUserName a {
 .contentPageMainColumn .jump_link {
     text-decoration: none;
 }
-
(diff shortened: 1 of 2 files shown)


DerekRoberts:
@cedar-technologies Thanks, looks good!  This code has been merged into a CI/CD branch below.  Please keep an eye on it and feel free to make further changes. :)

https://github.com/bcgov/gwells/pull/877

stephenhillier:
@cedar-technologies You're right about the main.css file being unwieldy to work with. For what it's worth, any new development with Vue uses Bootstrap4 and no `main.css`, we just aren't quite there yet with the existing well detail and well search templates.

We are trying out the new layout now. Thanks for submitting this!

PR #887, opened 2018-08-17
[1179] PATCH#2 - Pipeline Optimizations and Allocations

DerekRoberts:


Jenkinsfile
```diff
@@ -229,6 +229,7 @@ parallel (
             node('master'){
                 parallel (
                     "Load Fixtures": {
+                        sleep 10
```
Sybrand: @DerekRoberts We need to address this at some point with a wait check.

PR #888, opened 2018-08-22
[1235] Link wrap tweak

DerekRoberts:


DerekRoberts:
[EPIC#WellSummary | Links Not Word Wrapping Properly |0.5| !!!](https://trello.com/c/1PDp2mSu/1235-epicwellsummary-links-not-word-wrapping-properly-05)

app/backend/gwells/static/gwells/css/main.css
```diff
@@ -2300,7 +2300,7 @@ h5.twitterUserName a {
 #main-content #body a {
     word-wrap: break-word;
     /* RA-636 - Non standard for webkit */
-    word-break: break-all;
+    /* word-break: break-all; */
```
Sybrand: I'd prefer you remove the line entirely. Git will always have a record of what it used to be, we don't need dead code bloating the file. Also - the comment above it is also rendered redundant and you could remove that at the same time.

DerekRoberts: Done and squashed!

PR #892, opened 2018-08-23
[1238] PostGIS

DerekRoberts:


garywong-bc:
ooh I was so excited.. but don't see a #892 deployment on gwells-dev yet... once you have one I'd like to login, do some postgis commands and also some oracle fdw 'select from...'  thanks.  @DerekRoberts

DerekRoberts:
@garywong-bc I wish!  This was an investigation task, so anything outside of the most preliminary dabbling will be push into the next one.

PR #904, opened 2018-09-12
[1160] Submission step 13 (water quality)

stephenhillier:


app/frontend/src/submissions/views/SubmissionsHome.vue
```diff
@@ -254,11 +308,12 @@ export default {
       confirmSubmitModal: false,
       formSubmitLoading: false,
       formSubmitSuccess: false,
+      formSubmitError: false,
       saveFormSuccess: false,
       loadFormSuccess: false,
       confirmLoadModal: false,
       step: 1,
-      maxSteps: 11, // total number of wizard steps
+      maxSteps: 15, // total number of wizard steps
```
Sybrand: We need to think of a nice way to calculate this automatically based on the number of steps that have been added in as components - the solution we go for here is probably associated with the auto-numbering solution.

app/frontend/src/submissions/components/SubmissionForm/Step14Completion.vue
```diff
@@ -0,0 +1,162 @@
+<template>
+  <fieldset>
+    <legend>Step 14: Well Completion Data</legend>
```
Sybrand: I know we're going to address this as a whole in a separete task, but I've started dropping step numbers already in my code.

PR #920, opened 2018-09-26
Move keycloak idpHint value to environment variable.

NickPhura:
As discussed with Sybrand yesterday.
Add new config/environment variable to control keycloak hint.  
Allows alternatives to idir.

### DIFF
--- a/app/backend/gwells/views/api.py
+++ b/app/backend/gwells/views/api.py
@@ -26,6 +26,6 @@ class GeneralConfig(APIView):
         config = {
             "enable_data_entry": get_env_variable("ENABLE_DATA_ENTRY") == "True",
             "enable_google_analytics": get_env_variable("ENABLE_GOOGLE_ANALYTICS") == "True",
-            "sso_idp_hint": get_env_variable("SSO_IDP_HINT") or "idir"
+            "sso_idp_hint": get_env_variable("SSO_IDP_HINT", "idir")
         }
         return Response(config)
--- a/app/backend/gwells/views/api.py
+++ b/app/backend/gwells/views/api.py
@@ -26,6 +26,6 @@ class GeneralConfig(APIView):
         config = {
             "enable_data_entry": get_env_variable("ENABLE_DATA_ENTRY") == "True",
             "enable_google_analytics": get_env_variable("ENABLE_GOOGLE_ANALYTICS") == "True",
-            "sso_idp_hint": get_env_variable("SSO_IDP_HINT")
+            "sso_idp_hint": get_env_variable("SSO_IDP_HINT") or "idir"
         }
         return Response(config)
--- a/app/backend/gwells/views/api.py
+++ b/app/backend/gwells/views/api.py
@@ -25,6 +25,7 @@ class GeneralConfig(APIView):
     def get(self, request):
         config = {
             "enable_data_entry": get_env_variable("ENABLE_DATA_ENTRY") == "True",
-            "enable_google_analytics": get_env_variable("ENABLE_GOOGLE_ANALYTICS") == "True"
+            "enable_google_analytics": get_env_variable("ENABLE_GOOGLE_ANALYTICS") == "True",
+            "sso_idp_hint": get_env_variable("SSO_IDP_HINT")
         }
         return Response(config)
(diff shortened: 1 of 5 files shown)


NickPhura:
@DerekRoberts I didn't consider how this change might affect the pipeline.  We will at the very least need to include 1 more environment variable to account for the one added in the code.  Not sure if other changes would also be required.

app/backend/gwells/views/api.py
```diff
@@ -25,6 +25,7 @@ class GeneralConfig(APIView):
     def get(self, request):
         config = {
             "enable_data_entry": get_env_variable("ENABLE_DATA_ENTRY") == "True",
-            "enable_google_analytics": get_env_variable("ENABLE_GOOGLE_ANALYTICS") == "True"
+            "enable_google_analytics": get_env_variable("ENABLE_GOOGLE_ANALYTICS") == "True",
+            "sso_idp_hint": get_env_variable("SSO_IDP_HINT")
```
Sybrand: I think we should go for a default value of "idir" here for now. Since the only case where it's not idir, is for the purposes of functional testing, it's nice to have a default fallback.

e.g.: `get_env_variable("SSO_IDP_HINT", "idir")`

Opinions?

garywong-bc: definitely agree... IDIR should be the default.

Sybrand:
@NickPhura I've added changes to openshift/backend.dc.json - that should be all that's needed code wise for the pipeline. We need to set the value in openshift though - we need to discuss what it's going to be set where....

app/backend/gwells/views/api.py
DerekRoberts: IDIR seems like the only sane default.

DerekRoberts:
@stephenhillier IDIR isn't going to be our default?

DerekRoberts:
@NickPhura No worries about the pipeline just yet.  When all is decided we throw that in.

Sybrand:
@DerekRoberts once we add something other than IDIR, it makes sense to drop defaults entirely, or else we have to have a number of different login buttons with different defaults.

@stephenhillier we could drop IDIR as the default - if that' cool with Lindsay. We have to drop it at some point anyway.

NickPhura:
The main reason why I didn't include a default initially:

Calling get_env_variable("SSO_IDP_HINT") when the variable does not exist results in 'None'
Calling get_env_variable("SSO_IDP_HINT") when the variable does exist, but is set to "", results in 'None'

So the only way to unset a default value is to set SSO_IDP_HINT to some random string, which didn't seem intuitive.

Otherwise we have no way of knowing if the SSO_IDP_HINT variable is intentionally empty because we don't want to use the default or intentionally missing because we do want to use the default.

There might be some other workaround or alternative I haven't thought of.  Mostly its up to whatever the team prefers.

Sybrand:
A good solution might be a "keycloak" IDP_HINT we make, that pushes you to the keycloak login. Then we know any null value is incorrect. I'll see if I can figure out how to point to the keycloak login screen.

stephenhillier:
I think `None` will be represented in the API's json response as `null` and it may work fine to pass `idpHint: null` into the keycloak config.

app/backend/gwells/views/api.py
Sybrand: @NickPhura please change to default to "idir".

NickPhura:
Updated to default to "idir"

app/backend/gwells/views/api.py
```diff
@@ -25,6 +25,7 @@ class GeneralConfig(APIView):
     def get(self, request):
         config = {
             "enable_data_entry": get_env_variable("ENABLE_DATA_ENTRY") == "True",
-            "enable_google_analytics": get_env_variable("ENABLE_GOOGLE_ANALYTICS") == "True"
+            "enable_google_analytics": get_env_variable("ENABLE_GOOGLE_ANALYTICS") == "True",
+            "sso_idp_hint": get_env_variable("SSO_IDP_HINT") or "idir"
```
Sybrand: 2nd param of get_env_variable is the default value, so I'd prefer
```
"sso_idp_hint": get_env_variable("SSO_IDP_HINT", "idir")
```

PR #923, opened 2018-09-27
[1291] Docker Compose

DerekRoberts:


app/frontend/package.json
```diff
@@ -6,7 +6,7 @@
   "license": "Apache-2.0",
   "private": true,
   "scripts": {
-    "dev": "webpack-dev-server --inline --progress --config build/webpack.dev.conf.js",
+    "dev": "webpack-dev-server --inline --progress --config build/webpack.dev.conf.js --host 0.0.0.0",
```
stephenhillier: I would like to try getting this to work without this (I know this was my change but I don't think it's a good idea).

docker-compose.yml
```diff
@@ -0,0 +1,65 @@
...
+    command: /bin/bash -c "
+        set -x &&
+        cd /app/frontend &&
+        npm install &&
+        npm run build &&
+        npm run ${NPM_CMD:-dev}"
+    ports:
+      - "8080:8080"
+    volumes:
+      - ./app:/app
```
stephenhillier: i would still like node_modules to live on its own inside the container filesystem so that npm install doesn't affect the local `node_modules`. Some of them aren't cross platform.

DerekRoberts: @stephenhillier You'd made a good point about this earlier.  It's added back in.

app/frontend/package.json
DerekRoberts: @Sybrand Mind stepping in here?

Makefile
```diff
@@ -0,0 +1,24 @@
+
+################
+# General Jobs #
+################
+
+default: vue
+
+vue: prep
+	NPM_CMD=dev docker-compose up; docker-compose down
```
stephenhillier: We should hold off on adding docker-compose down. Right now anybody using this will likely need to add some test db data to develop against so we should let them clean up their containers explicitly when required.

DerekRoberts: @stephenhillier Db persists in ./.tmp.  I'll add a db clean to the Makefile in the AM.

DerekRoberts: @stephenhillier +fixtures?  What else?

stephenhillier:
Nice work, I left some comments on the commits (sorry... the proper review would have been more convenient).

Double check with Gary that none of the scripts are being referenced anywhere?

app/frontend/package.json
stephenhillier: another script like "npm run container" could use --host 0.0.0.0 so that "npm run dev" continues to bind to localhost only

rfb:
:+1:  - the only problem I encountered with the old docker-compose setup has been corrected

DerekRoberts:
[EPIC#Pipeline | Dev environment w/ Docker and Compose |2| !!](https://trello.com/c/YkH0Yo4F/1291-epicpipeline-dev-environment-w-docker-and-compose-2)

PR #940, opened 2018-10-09
task/1298_initialBDDStackUplift

NickPhura:
Trello: https://trello.com/c/gdefz3Xz/1298-epicregistry-write-authenticated-functional-tests-3

- Initial BDD uplift of pages and specs.
- Add new workflow specs.
- Add/update some IDs.
- Add apache license blurbs.

I fixed some of the issues from the comments on my last PR.  Mainly the broken unit test, and the apache license blurb.

At some point when we have all of the devs (and whomever else is interested) we can discuss which tests we like and which we don't, etc.

We can also do a general walk through with how I've chosen to organize things, and get feedback or suggestions.

@DerekRoberts [Users.groovy](https://github.com/bcgov/gwells/blob/19131fb9c0c7cf119c5afc2fd6ca3b4fa174377e/functional-tests/src/test/groovy/specs/traits/Users.groovy) has a number of username/password environment variables listed.  Locally, I have them set to the bdd test users I made in keycloak.  We'll need to do the same in Openshift, however we want to do that, when we get to the point where we want to add the functional tests to the pipeline.


functional-tests/src/test/groovy/specs/traits/Users.groovy
```diff
@@ -0,0 +1,37 @@
...
+
+package traits
+
+/**
+ * Methods to manage user credentials.
+ */
+trait Users {
+  Map env = System.getenv()
+  Map getUserOne() {
+    [username:env['GWELLS_USERNAME'], password:env['GWELLS_PASSWORD']]
```
NickPhura: This is a catch-all user with ALL roles and groups.  
This user should eventually not need to exist, but I ran into some issues where the "Registry" user didnt have all of the permissions i expected.  Maybe just need to double check how I set up the user in keycloak, or maybe there is an actual role/code issue. 

Can talk in person, when the time comes to try the tests in openshift.

cwevans:
[EPIC#Registry | Write authenticated functional tests |3| !!!](https://trello.com/c/gdefz3Xz/1298-epicregistry-write-authenticated-functional-tests-3)

PR #953, opened 2018-10-16
[1312] Tag Patch

DerekRoberts:


Jenkinsfile
```diff
@@ -798,6 +798,16 @@ for(String envKeyName: context.env.keySet() as String[]){
                     node('master') {
                         new OpenShiftHelper().deploy(this, context, envKeyName)
                         isDeployed = true
+
```
Sybrand: Is there a reason we're not using the openshift api, and instead are using bash scripts to call oc commands?

DerekRoberts: We're abandoning this pipeline and have decided not to polish it any further.  New pipeline soon.

PR #969, opened 2018-10-23
1297_NewFixtureData

NickPhura:
[Trello Card](https://trello.com/c/ioRPbYPZ/1297-epicsearch-review-functional-tests-5#)

(Ignore my comments from scrum this morning, I realized how I was overthinking this, so its no longer an issue)

Rename old registries.json, wellsearch.json.gz, and aquifers.json fixtures to registries_large.json, wellsearch_large.json.gz, and aquifers_large.json respectively.  
- We could delete these entirely, but they might be useful for some heavy-data tests in future?

Add new wellsearch.json fixture with reduced data/size.
- the new wellsearch.json just contains a small portion of the data from the original wellsearch.json (4 wells), which does cover the breadth of the tables from the original fixture data.
- Includes a small well tag id (2 - as requested by Lindsay) and an extra large id.

Add new registries.json fixture with reduced data/size.
- the new registries.json is brand new from users/companies created via the app, with a variety of statuses and types.

Add new aquifers.json fixture with reduced data/size (To match the reduced wellsearch.json fixture).
- the new aquifers.json contains a small portion of the data from the original aquifers.json.

(Not sure how well this code reviews..)


NickPhura:
[EPIC#Search | Review functional tests |5| !!!](https://trello.com/c/ioRPbYPZ/1297-epicsearch-review-functional-tests-5)

Sybrand:
@DerekRoberts @NickPhura  merging this in would break the pipeline - fixtures wouldn't get loaded. I can fix the jenkinsfile to pull in the renamed fixtures, but I don't want to cause conflicts on the pipeline you're working. Should we wait on merging this into 1.44.0 ?

DerekRoberts:
> @DerekRoberts @NickPhura merging this in would break the pipeline - fixtures wouldn't get loaded. I can fix the jenkinsfile to pull in the renamed fixtures, but I don't want to cause conflicts on the pipeline you're working. Should we wait on merging this into 1.44.0 ?

Good plan.  Let's aim for 1.44.0, when the timing isn't as close.

NickPhura:
@DerekRoberts @Sybrand 
Correct me if I am addressing the wrong issue:

In the jenkinsfile currently, it specifies "wellsearch.json.gz" but you don't need to specify the extensions at all (at least when running loaddata locally).  

You can just specify "wellsearch" and it will still scan and find the right fixture automatically.  

So we could update the jenkinsfile to just specify "wellsearch" and then it will work for the old zipped fixture AND the new updated unzipped fixture.

DerekRoberts:
@NickPhura Make that Jenkinsfile change and keep rolling!

NickPhura:
Pointed the Jenkinsfile to the original 'large' fixtures, for now.

PR #975, opened 2018-10-25
1336_BDD_AquiferData_View

NickPhura:
Code Changes/Updates
-------------------------------------------------------
- Add aquifers search geb page and bdd tests.
- Add ID's to real aquifers search page.
- Add .env support for the functional test gradle build.
- Trivial test comment update.

Trello
-------------------------------------------------------
https://trello.com/c/aT4I81oa/1336-epicaquiferdata-functional-tests-for-aquifer-data-view-2

Version
-------------------------------------------------------
Rebased on latest version of 'release/1.44.0' as of Nov 8, 2018.


NickPhura:
[EPIC#AquiferData | Functional Tests for Aquifer Data View |2| !!!](https://trello.com/c/aT4I81oa/1336-epicaquiferdata-functional-tests-for-aquifer-data-view-2)

app/frontend/src/aquifers/components/Search.vue
```diff
@@ -81,7 +84,7 @@ export default {
   },
   computed: {
     aquiferList () { return this.response && this.response.results },
-    emptyResults () { return this.response && this.response.count == 0 }
+    emptyResults () { return this.response && this.response.count === 0 }
```
NickPhura: This gave me a linting error, so I fixed it.

functional-tests/.env
```diff
@@ -1,5 +1,6 @@
-BASEURL=http://127.0.0.1:8000/
+BASE_URL=http://127.0.0.1:8000/gwells/
 GWELLS_USERNAME=bdduser1
 GWELLS_VIEWER_USERNAME=bdd-viewer
 GWELLS_SUBMISSION_USERNAME=bdd-submission
-GWELLS_REGISTRY_USERNAME=bdd-registry
\ No newline at end of file
+GWELLS_REGISTRY_USERNAME=bdd-registry
+GWELLS_PASSWORD=
```
NickPhura: I added a .env file, and a bit of code in build.gradle to read it and set the environment variables.

functional-tests/build.gradle
```diff
@@ -79,6 +79,21 @@ drivers.each { driver ->
...
+         * Set environment variables from a .env file, if it exists.
+         * Equals sign separated key value pairs, one per new line: SOME_KEY=some value
+         * These local environment variables take precedence over any environment variables set at the system level.
+         */
+        def envFile = file('.env')
+        if (envFile) {
+            envFile.readLines().each() {
+                def (key, value) = it.tokenize('=')
+                if (key && value) {
+                    environment key, value
```
NickPhura: This logic reads a .env file at this level and sets the environment variables.  Should make running the functional tests locally easier.

functional-tests/example.env
```diff
@@ -1,5 +1,6 @@
-BASEURL=http://127.0.0.1:8000/
+BASE_URL=http://127.0.0.1:8000/gwells/
 GWELLS_USERNAME=bdduser1
 GWELLS_VIEWER_USERNAME=bdd-viewer
 GWELLS_SUBMISSION_USERNAME=bdd-submission
-GWELLS_REGISTRY_USERNAME=bdd-registry
\ No newline at end of file
+GWELLS_REGISTRY_USERNAME=bdd-registry
+GWELLS_PASSWORD=
```
NickPhura: I included this example.env file, with the intention that you could rename it locally to .env and it would get used.

But maybe it shouldn't be committed at all?  I can always remove it and rebase so the history is wiped if needed.

Sybrand: I've been flip flopping on this. I've found it very useful to have .env files ready to go. This may become less of a concern if we all move over to using docker for our environments. Until then I'm ok with having a .env file - but I would drop the GWELLS_PASSWORD, just in case someone specifies something by accident and then commits it.

functional-tests/build.gradle
Sybrand: I would prefer it read form the environment itself only, and not a .env file. I just think it may be confusing, you'd have to have read this code, or some documentation to know which takes precedence.
Again - if we're all using docker, then the need for this code would go away.

functional-tests/example.env
NickPhura: Sure thats fair.  Removed.

I think long term, if we get docker to run the entire app, then this file/logic can be removed.  But in the interim its been useful for me, and hopefully whoever else wants to run the tests.

NickPhura:
It looks like my trello ticket is in the Done column, but I don't see this code in either of the master or release/1.44.0 branches.  And the PR was still open.  So not sure about that :/

DerekRoberts:
@NickPhura No idea why it was set to 'done', but it's back to 'in progress' now.

PR #978, opened 2018-10-25
[358] Submission preview form

stephenhillier:


app/frontend/src/submissions/components/SubmissionForm/MethodOfDrilling.vue
```diff
@@ -57,8 +57,8 @@
             <b-form-radio-group v-model="wellOrientationInput"
                                 stacked
                                 name="wellOrientationRadio">
-              <b-form-radio value="VERTICAL">Vertical</b-form-radio>
-              <b-form-radio value="HORIZONTAL">Horizontal</b-form-radio>
+              <b-form-radio value="True">Vertical</b-form-radio>
```
Sybrand: I though we had decided that using the codes makes more sense than using the boolean values?

app/frontend/src/submissions/components/SubmissionPreview/PreviewMap.vue
```diff
@@ -0,0 +1,88 @@
+<template>
```
Sybrand: You're going to need to pull in the esri map layers to be consistent with maps on submission and well detail.

app/frontend/package.json
```diff
@@ -28,6 +28,7 @@
     "vue-router": "^3.0.1",
     "vue-select": "^2.5.0",
     "vue-smoothscroll": "^0.1.1",
+    "vue2-leaflet": "^1.0.2",
     "vuex": "^3.0.1"
```
Sybrand: You're also going to need to pull in the esri leaflet package to load in the esri layers.

app/frontend/src/submissions/main.js
```diff
@@ -25,6 +25,8 @@ import FormInput from '@/common/components/FormInput.vue'
 import authenticate from '@/common/authenticate.js'
 import {FETCH_CONFIG} from '@/common/store/config.js'
 
+import 'leaflet/dist/leaflet.css'
```
Sybrand: Do we have to do this on main? Can it not be done on the component in question?

app/frontend/src/submissions/components/SubmissionPreview/SubmissionPreview.vue
```diff
@@ -0,0 +1,485 @@
...
+
+        // find the subclass to get its description
+        subclass = subclassCodes.find((code) => {
+          return code.well_subclass_guid === this.form.well_subclass
+        })['description']
+      }
+
+      return subclass
+    },
+    UTM () {
```
Sybrand: I think we should move this code into a common area - it's an almost exact copy of the same code in Coords.vue

stephenhillier:
[EPIC#Submission | Change submission prior to submitting |2| !!!](https://trello.com/c/eqC2twac/206-epicsubmission-change-submission-prior-to-submitting-2)

PR #983, opened 2018-10-27
[1362] Jenkins-to-GitHub Push Notifications

DerekRoberts:


DerekRoberts:
[EPIC#Pipeline | Notifications on GitHub |3| !!](https://trello.com/c/2bh0UKu7/1362-epicpipeline-notifications-on-github-3)

Jenkinsfile
```diff
@@ -35,12 +40,12 @@ pipeline {
     stage('ImageStreams ') {
       steps {
         script {
-          abortAllPreviousBuildInProgress(currentBuild)
```
stephenhillier: I think we should add this line back in and keep an eye on it. It was causing some pipelines to hang so I suggest:
```
echo "Cancelling previous builds..."
timeout(10) {
  abortAllPreviousBuildInProgress(currentBuild)
}
echo "Previous builds cancelled"

Just to make it more clear if the pipeline stops at this line.

Otherwise nice work, thanks for getting the utility functions in the repo.

DerekRoberts: Thanks and done!  (There are a few straggling parts for pulling the other repo, but it's coming along.)

PR #987, opened 2018-10-30
[1368] Changed lat/long to "text" input fields

Sybrand:
Due do the different way that browsers handle "number" fields, it's easier to just stick with a "text" type, and have Vue.js bind to text as a number.


Sybrand:
[EPIC#Submission | Map - decimal points on lat/long don't work on Firefox |1| !!!](https://trello.com/c/VIizUDYu/1368-epicsubmission-map-decimal-points-on-lat-long-dont-work-on-firefox-1)

PR #991, opened 2018-10-31
[1096] Edit page jump links

Sybrand:


app/frontend/src/submissions/components/SubmissionForm/Coords.vue
```diff
@@ -202,7 +202,11 @@ export default {
       type: Object,
       default: () => ({})
     },
-    errorCoordsNotProvided: Boolean
+    errorCoordsNotProvided: Boolean,
```
stephenhillier: `isInput: false`

app/frontend/src/submissions/components/SubmissionForm/ActivityType.vue
```diff
@@ -1,16 +1,15 @@
 <template>
   <fieldset>
-    <legend>Type of Work</legend>
+    <legend :id="id">Type of Work</legend>
```
stephenhillier: Probably a non-issue but I think the actual `id` text could have been put here instead of passed through from the parent ?

PR #993, opened 2018-10-31
[415] Create a new Aquifer

rfb:
- refactors aquifer form into a reusable component
- migrates Aquifer model to allow sequential id generation on create
- API exposes create for users with Aquifer Edit role
- links to Aquifer create from search home (needs work - where should this live?)

Needs tests - I want feedback on how it feels from @LindsayMacfarlane before writing tests.


rfb:
@garywong-bc can you check out this migration?

rfb:
[EPIC#AquiferData | Creating an Aquifer |8| !!!](https://trello.com/c/abx675cb/415-epicaquiferdata-creating-an-aquifer-8)

app/backend/aquifers/models.py
```diff
@@ -162,7 +162,7 @@ class Aquifer(AuditModel):
     This table holds ONLY the aquifers to which we have associated one or more wells.  It is not
     the definitive source of all aquifers in the province. 
     """
-    aquifer_id = models.PositiveIntegerField(
+    aquifer_id = models.AutoField(
```
garywong-bc: Yes, we need to set this to higher than the current highest (approx 1157) and add some for good measure (i.e. so that when we cut over to GWELLS we have left room).  

https://stackoverflow.com/a/50757423 seems to point to an additional migration step that ALTERS THE SEQUENCE #.. I'll ask @Sybrand  as he's the Django Migrate expert.

Sybrand: I think it's even simpler than that. See above (https://github.com/bcgov/gwells/pull/993/files#diff-bcddead7bbc7ec7a57ab59cc7f6b6622) ; you just make an empty migration, and add an operation using "migrations.RunSQL" - I've committed a piece of code to push up the sequence. Seems to work fine.

rfb:
@stephenhillier @Sybrand I feel like this is ready to go. I have the API tests (create/patch) added now.

rfb:
@Sybrand ok added auth blocks to that test - let me know how that goes!

PR #994, opened 2018-11-01
[1332] Bugfix/link to last known aquifer

garywong-bc:
Ensure that only the 'most recent aquifer edit/addition' is recorded against a well.


garywong-bc:
[EPIC#AquiferData | Get aquifer data into PROD |3| !!!](https://trello.com/c/S7o8zjfc/1332-epicaquiferdata-get-aquifer-data-into-prod-3)

app/database/scripts/wellsearch/wells_replication_stored_functions.sql
```diff
@@ -969,29 +969,36 @@ BEGIN
...
   raise notice '...aquifer data updated from mapping spreadsheet';
 
+-- With multiple aquifers per well, we choose the 'last updated' one 
+  WITH best_match AS (
+    SELECT DISTINCT ON (well.well_id)
+    well.well_tag_number,aws.aquifer_id  
+    FROM  wells.gw_aquifer_wells aws,
+          wells.wells_wells well
+    WHERE well.well_id = aws.well_id
+    ORDER BY well.well_id, coalesce(aws.when_updated, aws.when_created) DESC
```
Sybrand: Do all the records have either a when_updated or when_created? If for whatever reasons they're null anywhere, we could always order by the gw_aquifer_wells id to try and get the latest - trusting that the id's are created in order. Of course, if they date values are always populated this isn't a concern.

garywong-bc: When_created is guaranteed to have a value, hence my use of the coalesce() function on the order by.

garywong-bc: See my other comment about when_created bring mandatory. Oh also note that the aquifer_well table will be removed once @rfb 's code is merged in.

PR #998, opened 2018-11-02
[1104] Re-factor: form and preview split apart

Sybrand:
# Re-factor

@stephenhillier I did a pretty big re-factor as a 1st step before starting edit. I ripped out everything specific to the submission form into it's own component. I think this is a cleaner/better approach. It has some downsides, but I think it's better.

We should merge this bit before I finish the rest, otherwise merging in your changes is going to become very difficult.

# The good:
- More focused objects
- This approach simplifies things a bit - we don't have as many if this and that and the otherthing and not this thing, e.g.
```
<b-row v-if="!preview && activityType === 'STAFF_EDIT'">
```
becomes:
<b-row v-if="activityType === 'STAFF_EDIT'">

# The bad:
- More events, e.g:
```
          v-on:preview="handlePreviewButton"
          v-on:resetForm="resetForm"
```
- Thing happening all over the place.

### DIFF
--- a/app/frontend/src/submissions/components/SubmissionForm/ActivitySubmissionForm.vue
+++ b/app/frontend/src/submissions/components/SubmissionForm/ActivitySubmissionForm.vue
@@ -1,0 +1,474 @@
+<template>
+  <div>
+    <h1 class="card-title">
+      <b-row>
+        <b-col cols="12">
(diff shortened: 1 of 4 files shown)


Sybrand:
[EPIC#Edit | Person responsible |5| !!! [BV: 35]](https://trello.com/c/87NNMXRz/1104-epicedit-person-responsible-5-bv-35)

PR #1002, opened 2018-11-06
update Keycloak tokens

stephenhillier:
Added two calls to the Keycloak `updateToken()` method:
* on the `onTokenExpired` event, allowing access tokens to be continually refreshed if user leaves the page open for a long period.
* when loading previously saved tokens from localStorage (such as when returning to the site or refreshing the page).

I believe refresh tokens are valid for 10h.

### DIFF
--- a/app/frontend/src/common/authenticate.js
+++ b/app/frontend/src/common/authenticate.js
@@ -21,6 +21,19 @@ export default {
     return Vue.prototype.$keycloak
   },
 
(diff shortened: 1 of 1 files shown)


stephenhillier:
just a small note: `updateToken(60)` means a new token will only be requested if there is less than 60 seconds until the current token expires. However, the promise otherwise still resolves so the `.then()` block still fires.

PR #1017, opened 2018-11-10
[1397] Increased width of survey

Sybrand:


app/frontend/src/aquifers/components/Search.vue
```diff
@@ -18,7 +18,7 @@
     <b-alert
         show
         variant="info"
-        class="container mb-3"
+        class="container mb-3 col-sm-12"
```
stephenhillier:
```suggestion
        class="mb-3"
```

There's already a container wrapping this component

PR #1025, opened 2018-11-16
[1103] Submissions - Staff edit well owner

stephenhillier:
Frontend work:

* Added screens for editing well owner information and method of drilling.

Backend:

* went through models and added both null=True and blank=True to all char fields that had one or the other already set. This makes the behavior of these fields more consistent. Previously these fields were stored as a mix of either null or empty string.
* added fields owner_tel and owner_email


stephenhillier:
[EPIC#Edit | Method of Drilling |1| !!! [BV: 35]](https://trello.com/c/YJNGrU6n/1153-epicedit-method-of-drilling-1-bv-35)

stephenhillier:
[EPIC#Edit | Owner information |2| !!! [BV: 35]](https://trello.com/c/1XpYAJrV/1103-epicedit-owner-information-2-bv-35)

app/backend/wells/stack.py
```diff
@@ -166,7 +166,8 @@ def _stack(self, records, well: Well) -> Well:
             serializer = submissions.serializers.WellSubmissionStackerSerializer(submission)
             for source_key, value in serializer.data.items():
                 # We only consider items with values, and keys that are in our target
-                if value:
+                # an exception is STAFF_EDIT submissions (we need to be able to accept empty values)
+                if value or value == "" or value is False:
```
Sybrand: Does value == "" not mess up other submissions? Like, if an alteration follows a construction, with a blank string, would the alteration not override what was in the construction?

PR #1027, opened 2018-11-20
Enable USE_X_FORWARDED_HOST if SECURE_SSL_REDIRECT

rfb:
This _should_ fix a problem where the application reports back the deployed hostname (gwells-prod.pathfinder.gov.bc.ca) when it should be using the host name of the proxy (apps.nrs.gov.bc.ca).

We can see the problem on the swagger api page: https://apps.nrs.gov.bc.ca/gwells/api/#tag/wells

If you expand the "[GET] /wells/" dropdown, it displays

Server URL: http://gwells-prod.pathfinder.gov.bc.ca/gwells/api/v1/wells/

It should respond with:

Server URL: http://app.nrs.gov.bc.ca/gwells/api/v1/wells/

### DIFF
--- a/app/backend/gwells/settings/__init__.py
+++ b/app/backend/gwells/settings/__init__.py
@@ -307,3 +307,6 @@ class DisableMigrations(object):
         return None
 if get_env_variable('DISABLE_MIGRATIONS', None, strict=False, warn=False) == 'DISABLE_MIGRATIONS':
     MIGRATION_MODULES = DisableMigrations()
+
+SECURE_SSL_REDIRECT = bool(int(os.environ.get('SECURE_SSL_REDIRECT', 0)))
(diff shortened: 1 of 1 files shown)


Sybrand:
@stephenhillier I wonder if this is in any way related to the errors we were getting on leaflet?

rfb:
FYI - I lifted this snippit from the django-rest-swagger example app

https://github.com/marcgibbons/django-rest-swagger/blob/412f986bb5b8000681f9dcf544685908b386c364/example_app/tutorial/settings.py#L139

DerekRoberts:
@Sybrand `SECURE_SSL_REDIRECT = bool(int(os.environ.get('SECURE_SSL_REDIRECT', 0)))`?

PR #1031, opened 2018-11-20
[1402] data dump

Sybrand:


app/backend/wells/management/commands/export.py
```diff
@@ -0,0 +1,221 @@
...
+
+            # Write the values
+            row_index = 0
+            for row, record in enumerate(cursor.fetchall()):
+                values = []
+                num_values = 0
+                for col, value in enumerate(record):
+                    if value:
+                        num_values += 1
+                    values.append(value)
```
stephenhillier: Take care not to miss `orientation=False` (or other legit boolean fields). Is num_values just a count of "real" values?

For what it's worth I started thinking about fixing the well_orientation=False issue... sounds like it is NOT stored this way in the legacy database so we are free to change it on our end without breaking replication.

app/backend/wells/management/commands/export.py
```diff
@@ -0,0 +1,221 @@
...
+                if num_values > 1:
+                    # We always have a well_tag_number, but if that's all we have, then just skip this record
+                    row_index += 1
+                    for col, value in enumerate(values):
+                        if value:
+                            worksheet.write(row_index, col, value)
+                    csvwriter.writerow(values)
+        gwells_zip.write(csv_file)
+        if os.path.exists(csv_file):
+            os.remove(csv_file)
```
stephenhillier: is this because gwells_zip.write creates a different file?

app/backend/wells/management/commands/export.py
Sybrand: Good catch - will change to be explicit on "" and None

app/backend/wells/management/commands/export.py
Sybrand: Yep - once the csv is thrown into the zip file, we can get rid of it.

PR #1046, opened 2018-12-01
[1419] Smoke Tests

DerekRoberts:


Jenkinsfile
```diff
@@ -759,6 +759,69 @@ pipeline {
             }
         }
 
+
+        // Functional tests using BDD Stack
+        // See https://github.com/BCDevOps/BDDStack
+        stage('TEST - Smoke Tests') {
+            when {
+                expression { env.CHANGE_TARGET == 'master' }
```
Sybrand: I think we should run this in test and in production. I realize the smoke tests run as part of the normal tests, so that would mean any tests we run here, would run twice - but any issues in this section of Jenkins would then be exposed during test.

Jenkinsfile
```diff
@@ -759,6 +759,69 @@ pipeline {
...
+                                )
+                            ]
+                        ) {
+                            node("bddstack-${PR_NUM}") {
+                                //the checkout is mandatory, otherwise functional tests would fail
+                                echo "checking out source"
+                                checkout scm
+                                dir('functional-tests') {
+                                    try {
+                                        sh './gradlew -DchromeHeadlessTest.single=WellDetails chromeHeadlessTest'
```
Sybrand: I don't think this test exists any more?

> ./gradlew -DchromeHeadlessTest.single=WellDetails chromeHeadlessTest
> 
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> Execution failed for task ':chromeHeadlessTest'.
> > Could not find matching test for pattern: WellDetails


This might be a good alternative? (It does more, but it's all non-destructive)

`./gradlew -DchromeHeadlessTest.single=SearchSpecs chromeHeadlessTest`

Jenkinsfile
DerekRoberts: Not a bad idea!  Will make the addition.

Jenkinsfile
DerekRoberts: Also done!

DerekRoberts:
[EPIC#Pipeline | Re-Add Smoke Tests |3| !!](https://trello.com/c/IOdnMiAv/1419-epicpipeline-re-add-smoke-tests-3)

DerekRoberts:
[EPIC#Pipeline | Troubleshoot Local Storage in BDD |5| !!](https://trello.com/c/j9PeSDV6/1511-epicpipeline-troubleshoot-local-storage-in-bdd-5)

PR #1050, opened 2018-12-04
[1105] Edit Well Location

Sybrand:
Tests added
Map coordinate forced to numbers


app/frontend/src/submissions/components/SubmissionForm/CoordsMap.vue
```diff
@@ -37,6 +38,11 @@ export default {
       map: null
     }
   },
+  beforeCreate () {
+    // If lat/long aren't numbers, try to make them numbers.
+    this.$options.propsData['latitude'] = Number(this.$options.propsData['latitude'])
+    this.$options.propsData['longitude'] = Number(this.$options.propsData['longitude'])
+  },
```
stephenhillier: could this be in the parent component instead? 
```<coords-map :latitude="Number(latitude)">```

Sybrand: That's only a million times better!

Sybrand: Yes -  that's 10^6 time better.

garywong-bc:
FYI in a subsequent phase we will be loading into well.geom which will be a PostGIS spatial column.

PR #1052, opened 2018-12-05
[1106 ] Coordinate_aquisition_code and tests

Sybrand:
Adding coordinate_aquisition_code (to edit only, default to H in construction) and tests
Added post_migrate step to keep well_tag_number_sequence up to date
Added configurable random delay to requests
Handle delayed response from api/v1/wells
Added "loading" display on submission screen for well edit
Fixed marker bug (don't plot marker if map is not yet loaded)


Sybrand:
[EPIC#Edit | Geographic coordinates |2| !!! [BV: 35]](https://trello.com/c/zF6cyELf/1106-epicedit-geographic-coordinates-2-bv-35)

app/backend/submissions/serializers.py
```diff
@@ -167,13 +168,26 @@ class Meta:
         fields = '__all__'
 
 
+class CoordinateAquisitionCodeSerializer(serializers.ModelSerializer):
+    """ Serializes coordinate aquisition codes """
```
stephenhillier: I think this is a typo... should be acquisition

app/backend/gwells/middleware/GWellsRequestParsingMiddleware.py
```diff
@@ -1,11 +1,36 @@
...
+        """
+        As a developer, one is blind to various issues that may crop up due to network latency.
+        Set SIMULATE_NETWORK_LATENCY=True in your environment to introduce all manner of delays to frustrate
+        you.
+        """
+        SIMULATE_NETWORK_LATENCY = get_env_variable('SIMULATE_NETWORK_LATENCY',
+                                                    default_value=False,
+                                                    strict=False,
+                                                    warn=False)
+        if SIMULATE_NETWORK_LATENCY == 'True':
```
stephenhillier: minor quibble but if this is a long-term addition, can SIMULATE_NETWORK_LATENCY live in the `gwells.settings` file as
```
        SIMULATE_NETWORK_LATENCY = get_env_variable('SIMULATE_NETWORK_LATENCY',
                                                    default_value=False,
                                                    strict=False,
                                                    warn=False) == 'True'
```
and import the settings.SIMULATE_NETWORK_LATENCY boolean here?

stephenhillier: You could make it it's own middleware and then in the settings.py file do:
```
if SIMULATE_NETWORK_LATENCY:
    MIDDLEWARE = MIDDLEWARE + [ NetworkDelayMiddleware ] 
```
that way the network delay middleware will not be included in the middleware chain for requests on prod.

Semi-related note...it would be good to remove GwellsRequestParsingMiddleware.

Sybrand: I've created a ticket for GwellsRequestParsingMiddleware.

Delays should not be in the code. My solution, using nginx as reverse_proxy on local machine with a random delay.

brew install nginx-full --with-echo-module --with-lua-module
edit /usr/local/etc/nginx/nginx.conf ; nginx listening on port 8000, forwarding traffic to gwells running on 8001, with a random delay on the api:
`
        location /gwells/api/ {
           set $delay 0;
           rewrite_by_lua_block {
              ngx.var.delay = math.random(0, 5)
           }
           echo_sleep $delay;
           echo_exec @api;
        }

        location / {
           proxy_pass http://127.0.0.1:8001;
        }

        location @api {
            proxy_pass http://127.0.0.1:8001;
        }
`

Sybrand:
Used git push --force-with-lease - which somehow didn't take

PR #1056, opened 2018-12-07
Re-pointing to bcgov common image.

garywong-bc:


garywong-bc:
Let's leave this PR open for now.. I have a sinking suspicion that the bcgov\postgresql-96 image is NOT one built from bcgov\oracle-openshift_fdw master...  @DerekRoberts

PR #1062, opened 2018-12-11
IGNORE Refactoring DC to refer to pre-built bgov image. 

garywong-bc:


openshift/postgresql.dc.json
```diff
@@ -59,12 +59,24 @@
             "value": "1Gi"
         },
         {
-            "description": "Version of PostgreSQL image to be used (9.2, 9.4, 9.5 or latest).",
+            "description": "Version of PostgreSQL image to be used (9.5 or 9.6).",
             "displayName": "Version of PostgreSQL Image",
             "name": "IMAGE_STREAM_VERSION",
             "required": true,
-            "value": "9.5"
+            "value": "9.6"
```
stephenhillier: this is the tag for the `IMAGE_STREAM_NAME` you are planning to use (postgresql-9.6-oracle-fdw), so it will need to be `v1-stable`.  You could also just always pass it in as a parameter but this default value of 9.6 won't work.

garywong-bc: thx for the catch.. I'm still a little bit confused as the Jenkinsfile seems to only use the IMAGE_STREAM_NAME environment variable in the *output* of the deployment (e.g. naming/tagging the deployed image).  I don't see how  changing this value would affect anything in the deployed container.  For example the IMAGE_STREAM_NAMESPACE is blank... how do any of the IMAGE_STREAM_* stuff affect the source of the postgresql.dc.json file??
/cc @stephenhillier

stephenhillier: I see `IMAGE_STREAM_NAME` set in Jenkinsfile (line 207):
```
"IMAGE_STREAM_NAME=gwells-postgresql-${DEV_SUFFIX}-${PR_NUM}",
```

It is consumed in the DeploymentConfig ImageChange trigger (postgresql.dc.json):

```
"triggers": [
                    {
                        "imageChangeParams": {
                            "automatic": true,
                            "containerNames": [
                                "postgresql"
                            ],
                            "from": {
                                "kind": "ImageStreamTag",
                                "name": "${IMAGE_STREAM_NAME}:${IMAGE_STREAM_VERSION}",
                                "namespace": "${IMAGE_STREAM_NAMESPACE}"
                            },
                            "lastTriggeredImage": ""
                        },
                        "type": "ImageChange"
                    },
```

It's a bit confusing but I think the trigger is used to override the image specified in the DeployConfig (which is `"image": " "`).  The docs say if the image in the ImageChange trigger differs from the deploy image, that causes the deployment to load the image in the trigger.  So basically, whatever you pass into `${IMAGE_STREAM_NAME}:${IMAGE_STREAM_VERSION},` will be deployed, but that needs to be a real image and tag.  The reason the comments say "9.6, 9.5, latest" is because the file was originally for the official OpenShift postgres image.

As for namespace="", the apply commands are being run in the context of the `DEV_PROJECT` namespace already so I think it just defaults to that, and then falls back on "openshift" (where some of the base images are kept)

openshift/postgresql.dc.json
```diff
@@ -59,12 +59,24 @@
...
+            "description": "Indicator to enable pgcrypto extension (provided out-of-the-box with PostgreSQL).",
+            "displayName": "Flag to enable pgcrypto PostgreSQL extension.",
+            "name": "PGCRYPTO_EXTENSION",
+            "value": "Y"
+        },
+        {
+            "description": "Indicator to enable postgis extension (from official PostgeSQL YUM repo).",
+            "displayName": "Flag to enable postgis PostgreSQL extension.",
+            "name": "POSTGIS_EXTENSION",
+            "value": "Y"
```
stephenhillier: will we need to recreate the database or can we add a migration `create extension postgis` and re-use the existing data pvc?

garywong-bc: We cannot update 9.5 -> 9.6 (see https://devopspathfinder.slack.com/archives/C0PJD4LBU/p1544130149022500?thread_ts=1544127638.018600&cid=C0PJD4LBU).  So yes we'll need to:
- bring up Caddy to intercept users
- scale down application server
- pg_dump (using @DerekRoberts 's fancy https://github.com/bcgov/gwells/blob/release/1.47.0/openshift/scripts/oc-dump.sh )
- merge https://github.com/bcgov/gwells/pull/1062 into Release branch
- deploy to DEV (which will work as it's a brand new DB each time)
- deploy TEST/STAGING (which will fail as it's persistent DB with v9.5)
- oc debug into TEST DB pod and `rm /userdata`
- let it deploy again into blank `/userdata` folder, upgrading to v9.6 with postgis 2.4
- pg_restore (using @DerekRoberts 's fancy https://github.com/bcgov/gwells/blob/release/1.47.0/openshift/scripts/oc-restore.sh)
- scale up application server
- disable Caddy 
- test on STAGING/TEST before trying on PROD

PR #1065, opened 2018-12-11
Fix empty string be interpreted as 0, causing plot of point at 0,0

Sybrand:


Sybrand:
[EPIC#Edit | Map |3| !!! [BV: 35]](https://trello.com/c/G26u2BPo/1152-epicedit-map-3-bv-35)

PR #1067, opened 2018-12-14
[1344] changes to token/auth flow

stephenhillier:
* make use of keycloak's sessions/cookies (to the https://sso-test domains) to get a new token transparently. Tokens are no longer put in local storage
* router will not proceed to guarded routes if keycloak isn't ready (still testing/refining)

### DIFF
--- a/app/frontend/src/common/authenticate.js
+++ b/app/frontend/src/common/authenticate.js
@@ -46,10 +46,8 @@ export default {
         resolve() // We've already authenticated, have a header, and we've not expired.
       } else {
         // Attempt to retrieve a stored token, this may avoid us having to refresh the page.
-        const expiry = localStorage.getItem('gwells-session-expiry')
-        const expiryDate = expiry ? new Date(expiry * 1000) : null
 
-        store.commit('SET_KEYCLOAK_LOADING', true)
+        store.commit('SET_KEYCLOAK_READY', false)
 
         instance.init({
           onLoad: 'check-sso',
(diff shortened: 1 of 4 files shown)


app/frontend/src/common/store/auth.js
```diff
@@ -1,18 +1,26 @@
 export const SET_KEYCLOAK = 'SET_KEYCLOAK'
+export const SET_KEYCLOAK_READY = 'SET_KEYCLOAK_READY'
```
Sybrand: You may not need this state anymore (see comments below re. authentication now resolving promise correctly).

app/frontend/src/common/authenticate.js
```diff
@@ -45,31 +46,40 @@ export default {
         resolve() // We've already authenticated, have a header, and we've not expired.
       } else {
         // Attempt to retrieve a stored token, this may avoid us having to refresh the page.
-        const token = localStorage.getItem('token')
-        const refreshToken = localStorage.getItem('refreshToken')
-        const idToken = localStorage.getItem('idToken')
+
+        store.commit('SET_KEYCLOAK_READY', false)
```
Sybrand: You may not need the SET_KEYCLOAK_READY state anymore (see comments below re. authentication now resolving promise correctly).

app/frontend/src/common/store/auth.js
stephenhillier: I'll probably scrap this branch if you've come up with another solution. We'll have to touch base on what is left to be done/ what problems we still need to solve.

app/frontend/src/submissions/router/index.js
```diff
@@ -40,3 +41,24 @@ export default new Router({
...
+
+    const waitForLogin = () => {
+      setTimeout(() => {
+        if (store.getters.keycloakReady && store.getters.keycloak && store.getters.keycloak.authenticated) next()
+        else if (store.getters.keycloakReady && store.getters.keycloak && !store.getters.keycloak.authenticated) next({ path: '/', replace: true })
+        else waitForLogin()
+      }, 50)
+    }
+
+    waitForLogin()
```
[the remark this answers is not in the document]
stephenhillier: This could result in authenticate being called twice. Isn't that a bit messy? (not saying the solution on this branch isn't messy too, of course)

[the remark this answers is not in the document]
Sybrand: Yep - agreed - we can drop authenticate in main.js for submissions if we have it on the router. Will have to relook it when we get to having a one page app - as discussed.

PR #1071, opened 2018-12-19
[1513] Private documents not displaying bug

Sybrand:
- Changed user details to only be updated when changed
- Updated keycloak to latest version
- Changed when keycloak is changed in store
- Changed where resolve() is called as part of auth


Sybrand:
[Epic#WellSummary | Private documents not displaying |0| !!!](https://trello.com/c/kuLyeY9S/1513-epic-the-perfect-bug-0)

PR #1078, opened 2018-12-28
[1344] schedule token renewal instead of setting onTokenExpired

stephenhillier:
onTokenExpired seems flaky - it doesn't seem to run every time the token expires (we may be using it incorrectly). As soon as the token expires without onTokenExpired executing, we are left with a dead token.

I've set a scheduled renewal of half of the token's lifespan (or at least 60s).

One issue: if keycloak is down or unreachable when we try to refresh the token, the user may be logged out, which is unnecessary if there's just a network/server issue for a minute or two. We could try to address this next.


stephenhillier:
[EPIC#Submission | Don't allow token expiry to interrupt long form submission |3| !!](https://trello.com/c/BAutTgC8/1344-epicsubmission-dont-allow-token-expiry-to-interrupt-long-form-submission-3)

app/frontend/src/common/authenticate.js
```diff
@@ -62,15 +62,28 @@ export default {
...
-        instance.clearToken()
-      })
-    }
+
+  scheduleRenewal (instance) {
+    const expiresAt = instance.tokenParsed.exp
+
+    // set delay to 50% of token's lifespan (in seconds), or at least 60 seconds
+    let delay = (expiresAt - (Date.now() / 1000)) * 0.5
+    delay = delay > 60 ? delay : 60
```
Sybrand:
```suggestion
    delay = Math.max(60, delay)
```

app/frontend/src/common/authenticate.js
```diff
@@ -62,15 +62,28 @@ export default {
...
+
+    setTimeout(() => {
+      this.renewToken(instance)
+    }, delay * 1000)
+  },
+
+  renewToken (instance) {
+    instance.updateToken(1800).success(() => {
```
Sybrand: Why 1800 seconds? If I understand correctly, this code says that if the token is expiring in 1800 seconds or less, then update it. We don't actually need to update it if it's that far off do we?

app/frontend/src/common/authenticate.js
```diff
@@ -62,15 +62,28 @@ export default {
...
+      this.setLocalToken(instance)
```
Sybrand: On updateToken success, it actually returns "refreshed" (true/false) - so I think we only actually need to call setLocalToken if that has happened.

PR #1080, opened 2018-12-29
[414] Associate a well with an aquifer (on staff edit page)

stephenhillier:


app/backend/submissions/serializers.py
```diff
@@ -486,6 +487,16 @@ class Meta:
             'internal_comments',
             'alternative_specs_submitted',
             'decommission_description_set',
+            'avi',
```
Sybrand: Is this the aquifer vulnerability index? Either way - this seems to be the only acronym we have. Should we not be consistent and call things what they are?

app/backend/wells/models.py
```diff
@@ -779,6 +779,36 @@ class Well(AuditModel):
     consultant_company = models.CharField(
         max_length=200, blank=True, null=True, verbose_name='Consultant Company')
 
+    # Aquifer related data
+    avi = models.DecimalField(
```
Sybrand: See comment re. acronym above.

app/backend/wells/migrations/0048_auto_20181231_2249.py
```diff
@@ -0,0 +1,113 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='activitysubmission',
+            name='analytic_solution_type',
+            field=models.DecimalField(blank=True, decimal_places=2, max_digits=5, null=True, verbose_name='Analytic Solution Type'),
+        ),
+        migrations.AddField(
+            model_name='activitysubmission',
+            name='avi',
```
Sybrand: See comment re. acronym above.

app/backend/wells/models.py
```diff
@@ -1162,6 +1196,36 @@ class ActivitySubmission(AuditModel):
     decommission_details = models.CharField(
         max_length=250, blank=True, null=True, verbose_name="Decommission Details")
 
+    # Aquifer related data
+    avi = models.DecimalField(
```
Sybrand: See comment re. acronym above.

app/backend/submissions/serializers.py
```diff
@@ -467,6 +467,7 @@ class Meta:
             'water_quality_colour',
             'water_quality_odour',
             'ems_id',
+            "aquifer",
```
Sybrand: Can we use single quotes here for consistency?

app/backend/wells/serializers.py
```diff
@@ -245,6 +245,7 @@ class Meta:
             "observation_well_number",
             "observation_well_status",
             "ems",
+            "aquifer",
```
Sybrand: We're all over the show with quotes - could we at least be consistent in the same class. Maybe part of a bigger discussion, but I'm partial to single quotes in general.

https://www.python.org/dev/peps/pep-0008/#string-quotes

app/frontend/src/submissions/components/SubmissionForm/AquiferData.vue
```diff
@@ -0,0 +1,249 @@
...
+            :errors="errors['transmissivity']"
+            :loaded="fieldsLoaded['transmissivity']"
+        ></form-input>
+      </b-col>
+    </b-row>
+    <b-row>
+      <b-col cols="12" md="4">
+        <form-input
+            id="hydraulicConductivityInput"
+            label="Hydraulic Cconductivity"
```
Sybrand:
```suggestion
            label="Hydraulic Conductivity"
```

app/backend/wells/serializers.py
DerekRoberts: @Sybrand I'm violently opposed to single quotes and think they're the devil.

PR #1082, opened 2018-12-31
Revised dump script to ignore DB-wide spatial_ref_sys table 

garywong-bc:
- enabled postGIS in Docker

### DIFF
--- a/openshift/scripts/oc-dump.sh
+++ b/openshift/scripts/oc-dump.sh
@@ -92,8 +92,9 @@ SAVE_PATH=$( dirname ${SAVE_TO} )
 mkdir -p ${SAVE_PATH}
 oc exec ${POD_DB} -n ${PROJECT} -- /bin/bash -c '\
   pg_dump -U ${POSTGRESQL_USER} -d ${POSTGRESQL_DATABASE} -Fc -f /tmp/'${SAVE_FILE}' \
-	--no-privileges --no-tablespaces --schema=public \
+	--no-privileges --no-tablespaces --schema=public --exclude-table=spatial_ref_sys \
 '
+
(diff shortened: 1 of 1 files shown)


openshift/scripts/oc-dump.sh
```diff
@@ -92,8 +92,9 @@ SAVE_PATH=$( dirname ${SAVE_TO} )
 mkdir -p ${SAVE_PATH}
 oc exec ${POD_DB} -n ${PROJECT} -- /bin/bash -c '\
   pg_dump -U ${POSTGRESQL_USER} -d ${POSTGRESQL_DATABASE} -Fc -f /tmp/'${SAVE_FILE}' \
-	--no-privileges --no-tablespaces --schema=public \
+	--no-privileges --no-tablespaces --schema=public --exclude-table=spatial_ref_sys \
```
garywong-bc: @DerekRoberts With the new PostGIS extension, we now have a 'system/DB wide table' in the PUBLIC schema.  We do NO want to dump/restore that with the GWELLS objects, so I'm excluding it from the dump.

garywong-bc:
@DerekRoberts okay I've tested both the dump and the restore..  we shouldn't run into that error upon restore again `ERROR:  permission denied for relation spatial_ref_sys`

Tested on my local machine against PR-1082 DB:
```
(.venv) [garywong@NCM050128 scripts (task/enable_docker_postgis)]$ ./oc-restore.sh moe-gwells-dev/gwells-pgsql-dev-pr-1082 z.sql.dump
garywong-bc
DROP TABLE
...
Name: gwells-pgsql-dev-pr-1082
Proj: moe-gwells-dev
DB:   gwells
Size: 576K
File: z.sql.dump
```

PR #1085, opened 2019-01-02
[1271] Observation Well Information

DerekRoberts:


app/backend/aquifers/migrations/0009_auto_20190108_1925.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('aquifers', '0008_alter_sequence'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='aquifer',
+            name='mapping_year',
+            field=models.PositiveIntegerField(blank=True, help_text='Use the following format: <YYYY>', null=True, validators=[django.core.validators.MinValueValidator(1990), django.core.validators.MaxValueValidator(2019)], verbose_name='Date of Mapping'),
```
Sybrand: I'm not too keen on those validators. Especially 2019 as an upper limit. We do have validation as a future story, so I would just drop those entirely right now. Also - is this part of the story?

This looks like a migration that was missed in another PR? I think we should just drop it from this one?

PR #1088, opened 2019-01-07
[1480] Added presigned URL endpoints for S3 uploading

sjrumsby:
Added end-points for generating a presigned PUT URL for document uploading. This will allow a frontend client to request a key for uploading a document, and then be able to upload the document directly to an S3 compatible source - removing the need for the Django application to act as a middleman and handle the files / uploading directly.

### DIFF
--- a/app/backend/gwells/documents.py
+++ b/app/backend/gwells/documents.py
@@ -57,13 +57,19 @@ class MinioClient():
                 'S3_PUBLIC_ACCESS_KEY', warn=False)
             self.public_secret_key = get_env_variable(
                 'S3_PUBLIC_SECRET_KEY', warn=False)
+            self.use_https = int(get_env_variable(
+                'S3_USE_HTTPS', 1, warn=False))
 
             self.public_client = Minio(
                 self.public_host,
                 access_key=self.public_access_key,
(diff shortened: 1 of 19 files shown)


app/backend/gwells/documents.py
```diff
@@ -76,7 +82,7 @@ def create_private_client(self):
             self.private_host,
             access_key=self.private_access_key,
             secret_key=self.private_secret_key,
-            secure=True
+            secure=self.use_https
```
Sybrand: Is there a reason we can't always have secure=True?

app/backend/wells/views.py
```diff
@@ -202,3 +202,23 @@ def get(self, request):
...
+
+
+class PreSignedDocumentKey(APIView):
+    """
+    Get a pre-signed document key to upload into an S3 compatible document store
+
+    post: obtain a URL that is pre-signed to allow client-side uploads
+    """
+
+    permission_classes = (DjangoModelPermissionsOrAnonReadOnly,)
```
Sybrand: We're moving away from model permissions. You could leverage WellsEditPermissions right now if you wanted to.

PR #1094, opened 2019-01-11
WebPack Warning - bump webpack-dev-server to >=3.1.11

DerekRoberts:


stephenhillier:
Does not work for me... I think webpack-dev-server 3.x requires upgrading to webpack 4.0 (not sure on that)

PR #1095, opened 2019-01-12
[1488] Submission Tweaks

DerekRoberts:


app/backend/registries/urls.py
```diff
@@ -73,7 +73,9 @@
         views.PersonNoteDetailView.as_view(), name='person-note-detail'),
 
     # Person endpoints (drillers, well installers, and other instances of Person model)
-    url(r'^api/v1/drillers/names/$',
+    # TODO: There's some confusion between drillers and persons. Sometimes we're looking only for drillers,
+    # sometimes we're actually looking for people (pump installers, drillers etc.)
+    url(r'^api/v1/persons/names/$',
```
stephenhillier: I agree with the comment here but this has been discussed and for whatever reason we've settled on `drillers`.  We also published the endpoint`api/v1/drillers/names` so we can't change it.

app/frontend/src/aquifers/components/View.vue
```diff
@@ -148,7 +148,7 @@ export default {
     ...mapActions('documentState', [
       'uploadFiles'
     ]),
-    handleSaveSuccess () {
+    handleSaveSuccess (data) {
```
stephenhillier: a fix for this will come in a separate PR, I suggest removing this change.

app/frontend/src/submissions/components/SubmissionForm/Casings.vue
```diff
@@ -70,7 +70,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
                     text-field="description"
                     :state="getCasingError(index).casing_code ? false : null">
                   <template slot="first">
-                    <option :value="null" disabled>Select a type</option>
+                    <option value="" enabled>Select a type</option>
```
stephenhillier: `enabled` is not necessary

app/frontend/src/submissions/components/SubmissionForm/PersonResponsible.vue
```diff
@@ -246,14 +251,15 @@ export default {
...
+      this.drillerNameInput = (this.personResponsible && this.drillerSameAsPersonResponsible) ? this.personResponsible.name : ''
     },
     drillerSameAsPersonResponsible (val) {
       // keep driller name disabled & set to "person responsible", or leave it enabled and blank
       this.drillerNameInput = (this.personResponsible && this.drillerSameAsPersonResponsible) ? this.personResponsible.name : ''
     }
   },
   created () {
-    ApiService.query(`drillers/names/`).then((response) => {
+    ApiService.query(`persons/names/`).then((response) => {
```
stephenhillier: I think we should revert this change (see above comment)

PR #1108, opened 2019-01-25
Milestone 2

sjrumsby:


app/frontend/src/aquifers/components/View.vue
```diff
@@ -51,6 +56,7 @@
         v-on:cancel="navigateToView"
         :fieldErrors="fieldErrors"
         :record="record"
+        :aquifer_files="aquifer_files"
```
stephenhillier: Use camelCase for any javascript vars or component properties.  It gets caught by eslint and breaks the syntax highlighting

sjrumsby: This should be resolved now - along with all the other eslint warnings.

app/frontend/src/aquifers/components/Documents.vue
```diff
@@ -18,14 +18,50 @@
...
+            <div v-if="error">
+              {{error}}
+            </div>
+            <ul v-else-if="files && files.private && files.private.length">
+              <li v-for="(file, index) in files.private" :key="index">
+                <a :href="file.url" :download="file.name" target="_blank">{{file.name}}</a>
+                <a class="fa fa-trash fa-lg"
+                  variant="primary"
+                  v-if="editMode"
+                  style="margin-left: .5em"
```
stephenhillier: You can add class `ml-1` (or a higher value) for increasing the left margin instead of using an inline style.

PR #1118, opened 2019-01-31
[WIP] Added additional well list filters.

cole:
This is aimed at the wrong branch, as well as being incomplete, but it's really just here to have holes poked in it 😂. I'm looking for any things I'm missing/doing wrong so far.

My next steps here would be to implement the remaining filters, then develop postman tests for them, at which point I could submit a useful PR.

### DIFF
--- a/app/backend/wells/views.py
+++ b/app/backend/wells/views.py
@@ -13,7 +13,7 @@
 """
 from urllib.parse import quote
 
-from django.db.models import Prefetch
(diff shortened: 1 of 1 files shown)


stephenhillier:
Looks good so far, using `method=` on the CharFilter fields is much better than what was there before 👍

cole:
Ok, thanks! I'll proceed with this approach then.

PR #1119, opened 2019-02-01
Move Unit Tests from Build to Deploy

garywong-bc:
- forcing use of postgis over sqlite (e.g. test_gwells DB using PostGIS), in all stages
- backend references new python backend image, one that include GEOS/GDAL/PROJ.4 external libraries
- moved `stage('DEV - Django Unit Tests')` from the Build stage (running in moe-gwells-tools) to *after* the Deploy Stage (running in moe-gwells-dev)
- removed `atomic = ..` statements in migrations/*.py that are no longer necessary now that sqlite is out
- added `pygdal` to requirements.txt, now that the dependent libraries are in the backend image
- switched all `models.py` to import from `django.contrib.gis.db` rather than `django.db`
- `docker-compose.yml` references postgis image and python (with GEOS/GDAL/PROJ.4) backend image (this is temporary until we can get a local Docker image ready, based on alpine or centos)


garywong-bc:
Closing and restarting to reset the recreation of ConfigMaps.

PR #1121, opened 2019-02-01
[1388] Submission Tweaks II

DerekRoberts:


app/backend/gwells/fixtures/wellsearch-codetables.json
```diff
@@ -1678,7 +1692,7 @@
             "update_user": "ETL_USER",
             "update_date": "2017-07-01T08:00:00Z",
             "description": "Unknown",
-            "display_order": 110,
+            "display_order": 120,
```
stephenhillier: I don't know if it really matters for the fixtures but we might as well leave the display_order for existing data alone and just use an unused number like 101 for the new fixture.

app/frontend/src/submissions/components/SubmissionForm/Liner.vue
```diff
@@ -68,7 +68,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
             id="linerFrom"
             label="Liner From"
             hint="ft (bgl)"
-            type="number"
+            type="text"
             v-model.number="linerFromInput"
```
stephenhillier: does this work without throwing a console error when starting to type a decimal number like `.1` (without a leading 0)? If not we might need to remove the `.number` part of `v-model.number` as well.

DerekRoberts:
[EPIC#Submission | Tweaks |8| !!!](https://trello.com/c/NRsmqsrI/1388-epicsubmission-tweaks-8)

PR #1123, opened 2019-02-02
[1576] Feature/gis backend with configmap

garywong-bc:
- Moved Django Unit Tests from Build to post-deploy
- enabled anonymous pull of moe-gwells-tools images

NOTE: Need to verify the `when {}` in `Jenkinsfile`


Jenkinsfile
```diff
@@ -493,7 +492,56 @@ pipeline {
             }
         }
 
+        // the Django Unit Tests stage runs backend unit tests using a test DB that is
+        // created and destroyed afterwards. 
+        stage('DEV - Django Unit Tests') {
+            when {
+                expression { env.CHANGE_TARGET != 'master' && env.CHANGE_TARGET != 'demo' }
```
stephenhillier: We need to make sure the Django tests are run in all pipelines.  Let's remove the `when { expression ... }` condition blocks from all the "DEV" stages so every PR gets a dev environment with the testing.

app/backend/submissions/models.py
```diff
@@ -14,12 +14,10 @@
 from decimal import Decimal
 import uuid
 
-from django.db import models
+from django.contrib.gis.db import models
 from django.core.validators import MinValueValidator
-
 from gwells.models import AuditModel
```
stephenhillier: AuditModel in `app/backend/gwells/models` will need to have the version of models from `django.contrib.gis.db` too, since it's a base class for our other models.

Jenkinsfile
garywong-bc: so keep `expression { env.CHANGE_TARGET != 'master' }` ? 
@stephenhillier

app/backend/submissions/models.py
garywong-bc: Agreed.. but I do see still a mix of `from django.db` and `from django.contrib.gis.db`.. mainly the auto-generated migration files.   We'll have to test to make sure that's okay..

Jenkinsfile
```diff
@@ -493,7 +492,56 @@ pipeline {
             }
         }
 
+        // the Django Unit Tests stage runs backend unit tests using a test DB that is
+        // created and destroyed afterwards. 
+        stage('DEV - Django Unit Tests') {
+            when {
+                expression { env.CHANGE_TARGET != 'master' }
```
garywong-bc: @stephenhillier   I've 'resolved' this issue but please double check...

app/backend/submissions/models.py
garywong-bc: done **https://github.com/bcgov/gwells/blob/87b6ed7920ada6858266d68f1d74aadbc0f061e0/app/backend/wells/models.py#L15**

docker-compose.yml
```diff
@@ -51,7 +51,7 @@ services:
       - ./app:/app
       - /app/frontend/node_modules/
   backend:
-    image: python:3.6
+    image: garywongbc/python-36-gdal:v1
```
DerekRoberts: We shouldn't be accepting images from personal repos.  Please include any builds in the GWells repo.

garywong-bc: okay.. Ill do that now.. are we okay with the other personal repo?
https://github.com/bcgov/gwells/blob/87b6ed7920ada6858266d68f1d74aadbc0f061e0/docker-compose.yml#L27

Jenkinsfile
stephenhillier: I think we need to remove `env.CHANGE_TARGET != 'master'` as well (the whole `when` block).  The unit tests need to run every time the image is rebuilt, so there shouldn't be any `when { }` conditions here

docker-compose.yml
DerekRoberts: That's also bad!  Please use a build config to keep everything within our project.  I'll throw the other into the Trello board, since it's an issue.

stephenhillier: Can you replace `image: garywongbc/python-36-gdal:v1` with a `build:` section?  That way it will build off the Dockerfile in the repo instead of pulling from dockerhub.

```
build:
  context: .
  dockerfile: ./docker-images/python-gdal/Dockerfile
```

Jenkinsfile
garywong-bc: in the "olden" days, master branch was only for TEST/PROD deploy, while everything else only deployed to DEV.   I haven't kept up with the pipeline changes so I may be mistaken.. but don't master changes only proceed once DEV/TEST pipelines have passed?  Are we running the complete unit tests (and autocreation of test_gwells DB) even in master?

garywong-bc: `env.CHANGE_TARGET` is used throughout (e.g. https://github.com/bcgov/gwells/blob/3a55ada221c095e1dac30dd8903d0af1501398bf/Jenkinsfile#L366) so if we decide to make the Unit Tests run for all `!= master` we need to change them all consistently.  I'm beginning to wonder if I went down the wrong path in moving Unit Tests post-deployment.   Can I set up con-call tomorrow (or even today if @Sybrand  is not too sick)?  This PR is getting too big.

DerekRoberts:
[EPIC#Submission | Jenkins and Build Configs for GDAL |2| !!!](https://trello.com/c/iTyyeSB4/917-epicsubmission-jenkins-and-build-configs-for-gdal-2)

docker-compose.yml
garywong-bc: Fixed the reference to garywongbc Dockerhub with the 'build' section.

Jenkinsfile
garywong-bc: done.. 'Django Unit Tests' will run in all DEV pipelines (but != master)

docker-compose.yml
garywong-bc: @DerekRoberts  please doublecheck this fix (now calls build from the Dockerfile that is now in the repo).

app/backend/aquifers/migrations/0006_auto_20181003_1820.py
```diff
@@ -7,7 +7,7 @@
 
 
 class Migration(migrations.Migration):
-    atomic = settings.DATABASES.get('default').get('engine') == 'django.db.backends.postgresql'
+    atomic = settings.DATABASES.get('default').get('engine') == 'django.contrib.gis.db.backends.postgis'
```
Sybrand: We can drop all this logic, and just have atomic as you've done with the other files.

app/backend/.envrc
```diff
@@ -23,7 +23,7 @@ export DATABASE_SERVICE_NAME=localhost
 export LOCALHOST_SERVICE_HOST=localhost
 export BASEURL=http://127.0.0.1:8000/
 export LOCALHOST_SERVICE_PORT=5432
-export DATABASE_ENGINE=postgresql
+export DATABASE_ENGINE=postgis
```
Sybrand: I think we should entirely drop this variable - we're only using one DB engine now.

app/backend/gwells/database.py
```diff
@@ -19,21 +19,19 @@
 
 
 engines = {
-    'sqlite': 'django.db.backends.sqlite3',
     'postgresql': 'django.db.backends.postgresql',
+    'postgis': 'django.contrib.gis.db.backends.postgis',
```
Sybrand: Realistically we're never going to switch DB again - so I would just drop the entire engines section.

app/backend/gwells/settings/__init__.py
```diff
@@ -65,6 +65,10 @@
 # Fixtures dirs
 FIXTURES_DIRS = [FIXTURES_DIR]
 
+# GeoDjango External Library Paths
+GDAL_LIBRARY_PATH = get_env_variable('GDAL_LIBRARY_PATH', '/usr/local/lib/libgdal.so')
```
Sybrand: Just a note - this causes all kinds of trouble the moment you try to run local. Once you introduce these variables, I think it pretty much forces you to only work in docker.

app/backend/aquifers/migrations/0006_auto_20181003_1820.py
garywong-bc: Agreed.. done.

docker-compose.yml
```diff
@@ -24,7 +24,7 @@ services:
       MINIO_SECRET_KEY: minio1234
     command: server --address ":9001" /data
   db:
-    image: postgres:9.6
+    image: mdillon/postgis:9.6
```
Sybrand: Steve and Derek also commented on this - could we have this be a build: - and not rely on personal repo?

PR #1127, opened 2019-02-06
[1550] Well Search API

cole:
Work in progress; more Postman tests to come.


cole:
@Sybrand, @stephenhillier, or @DerekRoberts (not sure who is best) if one of you has time on Monday can you take a look at the Postman tests I have so far and see if they're at the level you're expecting? I don't think I'll be around to discuss until Tuesday unfortunately.

Currently I have 61 test requests running as part of the suite. With min and max values for range filters it looks like I will need ~150 to cover all of the filters (success cases only), plus another ~220 or so to cover ordering. That is doable but seems like it would be a pain to maintain. I've looked into using data files to drive the collection run with multiple iterations, but it's still pretty clunky. Hopefully someone with more Postman experience can advise?

cole:
OK, based on slack discussions I've tried to achieve a good balance of testing the different types of filters in Postman, rather than each filter (since there are so many). I've also organized them into folders so they're a bit easier to follow. This PR should be ready to merge now — let me know if there's anything I'm missing.

DerekRoberts:
[EPIC#AdvancedSearch | Well Search API |8| !!!](https://trello.com/c/tZJBtHFs/1550-epicadvancedsearch-well-search-api-8)

PR #1132, opened 2019-02-09
[1557] Added Published / Unpublished state to wells

AlexZorkin:
- Can edit well_publication_status from well edit view
- Excludes wells with unpublished state from search and export. 
- Sql commands in export have been tested and work as expected, testing in live cron job environment is needed.


app/backend/wells/management/commands/export.py
```diff
@@ -195,27 +196,33 @@ def generate_files(self, zip_filename, spreadsheet_filename):
  lhc.lithology_hardness_code = lithology_description.lithology_hardness_code
  left join lithology_colour_code as lcc on
  lcc.lithology_colour_code = lithology_description.lithology_colour_code
- order by well_tag_number""")
+ left join well on well.well_tag_number = lithology_description.well_tag_number
```
Sybrand: I think this should be an inner join on the well table. We're not interested in lithology records for unpublished wells.

app/backend/wells/management/commands/export.py
```diff
@@ -195,27 +196,33 @@ def generate_files(self, zip_filename, spreadsheet_filename):
...
+ where well.well_publication_status_code = 'Published'
+ order by lithology_description.well_tag_number""")
         ########
         # CASING
         ########
-        casing_sql = ("""select well_tag_number, casing_from, casing_to, diameter, casing_code,
+        casing_sql = ("""select casing.well_tag_number, casing_from, casing_to, casing.diameter, casing_code,
  casing_material_code, wall_thickness, drive_shoe from casing
- order by well_tag_number""")
+ left join well on well.well_tag_number = casing.well_tag_number
```
Sybrand: Again, I think this should be an inner join.

app/backend/wells/management/commands/export.py
```diff
@@ -195,27 +196,33 @@ def generate_files(self, zip_filename, spreadsheet_filename):
...
+ where well.well_publication_status_code = 'Published'
+ order by casing.well_tag_number""")
         ########
         # SCREEN
         ########
-        screen_sql = ("""select well_tag_number, screen_from, screen_to, internal_diameter,
+        screen_sql = ("""select screen.well_tag_number, screen_from, screen_to, internal_diameter,
  screen_assembly_type_code, slot_size from screen
- order by well_tag_number""")
+ left join well on well.well_tag_number = screen.well_tag_number
```
Sybrand: Again, I think this should be an inner join.

app/backend/wells/management/commands/export.py
```diff
@@ -195,27 +196,33 @@ def generate_files(self, zip_filename, spreadsheet_filename):
...
         # PERFORATIONS
         ##############
-        perforation_sql = ("""select well_tag_number, liner_from, liner_to, liner_diameter,
- liner_perforation_from, liner_perforation_to, liner_thickness
- from
- perforation
- order by well_tag_number""")
+        perforation_sql = ("""select p.well_tag_number, p.liner_from, p.liner_to, p.liner_diameter,
+ liner_perforation_from, liner_perforation_to, p.liner_thickness from perforation as p
+ left join well on well.well_tag_number = p.well_tag_number
```
Sybrand: Again, I think this should be an inner join.

app/backend/wells/management/commands/export.py
```diff
@@ -175,17 +175,18 @@ def generate_files(self, zip_filename, spreadsheet_filename):
  registries_person.person_guid = well.person_responsible_guid
  left join registries_organization on
  registries_organization.org_guid = well.org_of_person_responsible_guid
+ where well.well_publication_status_code = 'Published'
```
Sybrand: None of the wells are actually going to have a "Published" status. They're all code to have Null for well_publication_status_code. So really the query has to be limit on "Published" or Null.

app/backend/wells/management/commands/export.py
```diff
@@ -195,27 +196,33 @@ def generate_files(self, zip_filename, spreadsheet_filename):
  lhc.lithology_hardness_code = lithology_description.lithology_hardness_code
  left join lithology_colour_code as lcc on
  lcc.lithology_colour_code = lithology_description.lithology_colour_code
- order by well_tag_number""")
+ left join well on well.well_tag_number = lithology_description.well_tag_number
+ where well.well_publication_status_code = 'Published'
```
Sybrand: See comment above re. well_publication_status_code being Null

app/backend/wells/management/commands/export.py
```diff
@@ -195,27 +196,33 @@ def generate_files(self, zip_filename, spreadsheet_filename):
...
+ order by lithology_description.well_tag_number""")
         ########
         # CASING
         ########
-        casing_sql = ("""select well_tag_number, casing_from, casing_to, diameter, casing_code,
+        casing_sql = ("""select casing.well_tag_number, casing_from, casing_to, casing.diameter, casing_code,
  casing_material_code, wall_thickness, drive_shoe from casing
- order by well_tag_number""")
+ left join well on well.well_tag_number = casing.well_tag_number
+ where well.well_publication_status_code = 'Published'
```
Sybrand: See comment above re. well_publication_status_code being Null

app/backend/wells/management/commands/export.py
```diff
@@ -195,27 +196,33 @@ def generate_files(self, zip_filename, spreadsheet_filename):
...
+ order by casing.well_tag_number""")
         ########
         # SCREEN
         ########
-        screen_sql = ("""select well_tag_number, screen_from, screen_to, internal_diameter,
+        screen_sql = ("""select screen.well_tag_number, screen_from, screen_to, internal_diameter,
  screen_assembly_type_code, slot_size from screen
- order by well_tag_number""")
+ left join well on well.well_tag_number = screen.well_tag_number
+ where well.well_publication_status_code = 'Published'
```
Sybrand: See comment above re. well_publication_status_code being Null

app/backend/wells/management/commands/export.py
```diff
@@ -195,27 +196,33 @@ def generate_files(self, zip_filename, spreadsheet_filename):
...
         ##############
-        perforation_sql = ("""select well_tag_number, liner_from, liner_to, liner_diameter,
- liner_perforation_from, liner_perforation_to, liner_thickness
- from
- perforation
- order by well_tag_number""")
+        perforation_sql = ("""select p.well_tag_number, p.liner_from, p.liner_to, p.liner_diameter,
+ liner_perforation_from, liner_perforation_to, p.liner_thickness from perforation as p
+ left join well on well.well_tag_number = p.well_tag_number
+ where well.well_publication_status_code = 'Published'
```
Sybrand: See comment above re. well_publication_status_code being Null

app/backend/wells/views.py
```diff
@@ -69,7 +69,7 @@ class WellDetail(RetrieveAPIView):
     """
     serializer_class = WellDetailSerializer
 
-    queryset = Well.objects.all()
+    queryset = Well.objects.exclude(well_publication_status='Unpublished')
```
Sybrand: A user editing a well still needs this endpoint to work - so either we have to no filter here, or only filter on users who don't have edit permissions. This endpoint is used in app/frontend/src/submissions/views/SubmissionsHome.vue for loading the form.

app/backend/wells/views.py
```diff
@@ -151,7 +151,7 @@ class WellListAPIView(ListAPIView):
 
     permission_classes = (DjangoModelPermissionsOrAnonReadOnly,)
     model = Well
-    queryset = Well.objects.all()
+    queryset = Well.objects.exclude(well_publication_status='Unpublished')
```
Sybrand: If a logged in user wants to edit a well, to make it published, the user would have to be able to find it. I think we should only apply this filter to users who aren't logged in with well edit rights.

app/backend/wells/views.py
```diff
@@ -185,7 +185,7 @@ class WellTagSearchAPIView(ListAPIView):
 
     permission_classes = (DjangoModelPermissionsOrAnonReadOnly,)
     model = Well
-    queryset = Well.objects.only('well_tag_number', 'owner_full_name').all()
+    queryset = Well.objects.exclude(well_publication_status='Unpublished').only('well_tag_number', 'owner_full_name')
     pagination_class = None
```
Sybrand: If a logged in user wants to edit a well, to make it published, the user would have to be able to find it. I think we should only apply this filter to users who aren't logged in with well edit rights.

app/backend/wells/models.py
```diff
@@ -418,6 +418,23 @@ def save(self, *args, **kwargs):
         super(WellStatusCode, self).save(*args, **kwargs)
 
 
+class WellPublicationStatusCode(AuditModel):
```
Sybrand: Can we please get all the columns null=False as suggested by Meherzad?

DerekRoberts:
[Help | Wells DB Change Broke Replication |2| !!!](https://trello.com/c/iQL93wLe/1616-help-wells-db-change-broke-replication-2)

garywong-bc:
Hi confirmed that this PR has the fixes to the Data Replication scripts.  The two steps were run manually on gwells-pgsql-dev-pr-1132-1-8mqzt  and completed sucessfully:
```
sh-4.2$ psql -t -d $POSTGRESQL_DATABASE -U $POSTGRESQL_USER -c 'SELECT db_replicate_step1(_subset_ind=>true);'
sh-4.2$ psql -t -d $POSTGRESQL_DATABASE -U $POSTGRESQL_USER -c 'SELECT db_replicate_step2();'
...
....
NOTICE:  Finished replicating WELLS to GWELLS.
```

PR #1149, opened 2019-02-13
Increased comments column to 2000 in xform_well, added unknown well class code in migration

Sybrand:


DerekRoberts:
[Help | Wells DB Change Broke Replication |2| !!!](https://trello.com/c/iQL93wLe/1616-help-wells-db-change-broke-replication-2)

PR #1152, opened 2019-02-14
[1606] Docker Compose - use default python, add GDAL

DerekRoberts:


DerekRoberts:
[EPIC#Pipeline | Create Backend Dockerfile and Use Only Available Images |5| !!](https://trello.com/c/APHGud91/1606-epicpipeline-create-backend-dockerfile-and-use-only-available-images-5)

PR #1155, opened 2019-02-16
Alternative docker-compose file to get around GDAL/GEOS issues.

garywong-bc:
Be sure to read the comments at top of the file, as you do need to be authenticated to pull from the protected image repo's.

@stephenhillier  You're much more familiar with Docker than I.. please review, especially the
``` 
volumes:
      - db-data:/var/lib/pgsql/data
```
part.  If it's not best practice to make it persistent, let me know.

### DIFF
--- a/docker-compose.openshift.yml
+++ b/docker-compose.openshift.yml
@@ -1,0 +1,137 @@
+version: '3'
+# NOTE Before running `docker-compose -f docker-compose.openshift.yml up`, you need to:
+#     1. Be logged into the BC Gov OpenShift Console 
+#          https://console.pathfinder.gov.bc.ca:8443/console/catalog)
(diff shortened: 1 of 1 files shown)


cole:
@garywong-bc I don't have an OpenShift login, and as I understand it there are no plans to get me one, so this doesn't work for my current development setup.

I was able to get the standard dockerfile working by adding the `python-gdal` package to the apt install section and _removing_ the GDAL requirement from requirements.txt (it seems that the pip GDAL package isn't backwards compatible, so 2.4 won't work with the 2.1 debian package). I'm not sure if that's helpful as an alternative, or if 2.4 is actually required? If it is helpful, I can submit a PR with the changes.

DerekRoberts:
@garywong-bc Thanks for the help - @cole got this solved!

PR #1157, opened 2019-02-18
[1606] Docker Compose and Dockerfiles

DerekRoberts:


DerekRoberts:
[EPIC#Submission | Jenkins and Build Configs for GDAL |2| !!!](https://trello.com/c/aObYTsI2/1576-epicsubmission-jenkins-and-build-configs-for-gdal-2)

DerekRoberts:
> this is a good start and I think it's a big step in the right direction but let's make sure it can work with the staging and prod environments, as well as locally. Right now one of the requirements for prod is GDAL python module in the requirements.txt file.

That was merged breaking local dev.  I'd be happy to support and fix it, but not in this PR.

PR #1162, opened 2019-02-23
Addition of geom and updates to ETL

garywong-bc:
- leverage PostGIS, GDAL, and GEOS,  adding a new `well.geom` attribute via GeoDjango
- leaving existing `well.longitude` and `well.latitude` numeric columns to follow our "expand-and-then-contract" pattern for renaming/dropping attributes, but update them via a migration step `app/backend/wells/migrations/0063_well_geom_convert.py`
- update Data Replication to populate `geom` column


app/backend/wells/migrations/0063_well_geom_convert.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
```
Sybrand: I think we can just skip this migration - fixtures are loaded after running migrations locally, so this is only useful for someone with an existing database, until they drop it - or for the time between this migration being run, and replication kicking in.
Developers could run this sql manually to fix up their data.
We're going to need to update the fixtures regardless at some point anyway.

garywong-bc: do you mean the file https://github.com/bcgov/gwells/blob/release/app/backend/gwells/fixtures/wells.json ? I'm able to update it to include the geom value, but I noticed that this file is not being used (and in fact has a conflict so perhaps it hasn't been kept up-to-date?).
```
django.db.utils.IntegrityError: Problem installing fixture '/Users/garywong/proj/gwells/app/backend/gwells/fixtures/wells.json': Could not load wells.Well(pk=110749): duplicate key value violates unique constraint "well_identification_plate_number_key"
DETAIL:  Key (identification_plate_number)=(99) already exists.
```

PR #1174, opened 2019-02-27
Fix Document View On Well Summary

AlexZorkin:


app/frontend/src/wells/components/Documents.vue
```diff
@@ -93,7 +96,7 @@ export default {
       if (this.well) {
         this.loadFiles()
       }
-    }
+    },
```
stephenhillier: minor, but this light up our IDE linting ("no trailing comma")

PR #1175, opened 2019-02-27
[1659] minor fixes for single page app

stephenhillier:
* Added a "back to top" floating button.  This will be added to the entire app.
* Adjusted the well location search (the API endpoint that populates the map markers)


stephenhillier:
I'd like to fix the views.py `count > 5000` thing in a separate PR, but I'd like to leave the code in for now because removing it makes the map impossible to load at higher zoom levels (too many markers).

PR #1176, opened 2019-02-28
[1674] Fix Well Tag Search

AlexZorkin:
- removed params from url that set ordering by well_tag_number. This was causing a static page to be returned. Well tag number is the default ordering for this query anyways.

### DIFF
--- a/app/frontend/src/submissions/store/index.js
+++ b/app/frontend/src/submissions/store/index.js
@@ -49,7 +49,7 @@ const submissionStore = {
     [FETCH_WELLS] ({ commit }) {
       if (!this.state.wells) {
         // fetch the wells once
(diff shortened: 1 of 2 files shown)


app/frontend/src/submissions/store/index.js
```diff
@@ -49,7 +49,7 @@ const submissionStore = {
     [FETCH_WELLS] ({ commit }) {
       if (!this.state.wells) {
         // fetch the wells once
-        ApiService.query('wells/tags/?ordering=well_tag_number').then((res) => {
+        ApiService.query('wells/tags').then((res) => {
```
Sybrand:
```suggestion
        ApiService.query('wells/tags?ordering=well_tag_number').then((res) => {
```

PR #1179, opened 2019-03-01
[1330] Align Django groups with Keycloak roles

Sybrand:
- Squashed migrations on gwells as part of removing migration step that loads various group permissions
- Removed registries_adjudicator and registries_statutory_authority (now relying on finer grainer roles)
- Removed model based permissions from registries
- Removed all unused roles
- Squashed registries migrations


Sybrand:
[EPIC#Login | Align Django groups with Keycloak roles and drop reliance on Django Permissions |3| !!](https://trello.com/c/vhjiDvYC/1330-epiclogin-align-django-groups-with-keycloak-roles-and-drop-reliance-on-django-permissions-3)

PR #1180, opened 2019-03-01
[1545] Errors when email not specified in SSO

Sybrand:
- Changed code so it will no longer cause errors if email is not specified in SSO.
- Cleaned up gwells.Profile (removed un-used fields).
- Fixed some linting errors.


Sybrand:
[EPIC#Edit | API fails when no email specified for user |1| !!](https://trello.com/c/IVQbCt86/1545-epicedit-api-fails-when-no-email-specified-for-user-1)

stephenhillier:
Not sure how to test with BCeID but I did some manual testing with my IDIR account and everything looks good so far.

PR #1181, opened 2019-03-02
[1546] Using province border geometry to establish if point is inside bc

Sybrand:


Sybrand:
[EPIC#Submission | Tweaks - Restrict map to BC with validation messages (reduces time required to identify and correct mistakes) |5| !!!](https://trello.com/c/pME3kG9K/1546-epicsubmission-tweaks-restrict-map-to-bc-with-validation-messages-reduces-time-required-to-identify-and-correct-mistakes-5)

PR #1209, opened 2019-03-08
WIP:  Bootstrap 4.1 update

stephenhillier:
* updated bootstrap-vue
* updated our import of bcgov/bootstrap-theme (we may want to pin this onto a certain commit)

Extra fixes:
* Added flex layout for main App.vue template, because recent updates to the bcgov/bootstrap-theme seemed to change the behavior of our previous header/main/footer layout.  Needs testing in Internet Explorer

Still needs fixing:
* Registries page container is for some reason constrained to 745px (very small).
* IE11 testing
* Search/select boxes (like "Person responsible for drilling") are slightly larger than other input boxes now


app/frontend/src/common/common.scss
```diff
@@ -56,12 +60,12 @@ footer {
   font-weight: 500!important;
 }
 
-// BC Gov Bootstrap 4 theme override
-.custom-select[size]:not([size="0"]) {
-  height:auto;
-  padding-right:.75rem;
-  background-image:none;
-}
+// // BC Gov Bootstrap 4 theme override
```
Sybrand: Might as well just remove this commented code?

app/frontend/src/common/common.scss
```diff
@@ -31,7 +35,7 @@
 }
 
 html {
-  font-size: 14px;
+  font-size: 14px!important;
   background-color: #f1f1f1
```
Sybrand:
```suggestion
  background-color: #f1f1f1;
```

app/frontend/package.json
```diff
@@ -17,9 +17,10 @@
     "watch": "./node_modules/.bin/webpack --config build/webpack.dev-watch.conf.js --watch --display-entrypoints"
   },
   "dependencies": {
+    "@bcgov/bootstrap-theme": "git+https://github.com/bcgov/bootstrap-theme.git",
```
Sybrand: I wonder if it's work specifying the latest commit? They aren't using any tags in that repo, and we probably don't want to be caught unawares?
```suggestion
    "@bcgov/bootstrap-theme": "git+https://github.com/bcgov/bootstrap-theme.git#7809a4fae4f2e44686bda62df17540e487fa5413",
```

stephenhillier:
Still need to address one thing:  Registries page is very narrow

stephenhillier:
note: registries width is fixed, but the header/main/footer layout does not work in IE11.

PR #1211, opened 2019-03-12
Fix for Pending Trigger Events Registries Migration

AlexZorkin:


AlexZorkin:
Tested against production database and was working.

PR #1222, opened 2019-03-15
[1625] Added Lag Input for Latitude Longitude Fields on Submission and Edit

AlexZorkin:


app/frontend/src/submissions/components/SubmissionForm/Coords.vue
```diff
@@ -295,7 +295,8 @@ export default {
       latitudeDMSValidation: false,
       coordinateLookup: new Map(),
       coordinateResolveLookup: new Map(),
-      validCoordinate: null
+      validCoordinate: null,
+      timeout: null
```
stephenhillier: a name like `coordCheckTimeout` or `coordCheckDebounce` or something similar might be helpful

PR #1234, opened 2019-03-20
adjust match_any behavior for coord bounds (prevent OR+AND on the same geom column)

stephenhillier:


cole:
It seems like this still might be slow — the two points approach was a bad idea. I've submitted #1236 as an alternate.

PR #1235, opened 2019-03-21
[1713] Well search fixes

stephenhillier:
* tie search results table to the current bounds of the map
* fix container width


cole:
I think this still might be slow — the way I implemented the bounding box as two points is fundamentally not great, I think. I'll submit a PR to revise and you can decide.

PR #1243, opened 2019-03-22
Minor well search page fixes

stephenhillier:


app/frontend/src/wells/views/WellSearch.vue
```diff
@@ -446,15 +456,39 @@ export default {
...
         return []
+      }).finally(() => {
+        this.pendingSearches.shift()
       })
     },
+    handleScroll () {
+      const pos = this.$el.querySelector('#map').scrollTop | 100
+      this.scrolled = window.scrollY > 0.9 * pos
+    },
+    cancelWellSearches () {
```
AlexZorkin: This method will cancel only half of the pendingSearches because it is iterating forward while popping the last value in the array. Iterating in reverse would cancel all pending searches.
for (let i = this.pendingSearches.length -1; i >= 0; i--) {
        const req = this.pendingSearches.pop()
        req.cancel()
      }

app/frontend/src/wells/views/WellSearch.vue
```diff
@@ -446,15 +456,39 @@ export default {
...
+      for (let i = 0; i < this.pendingSearches.length; i++) {
+        const req = this.pendingSearches.pop()
+        req.cancel()
+      }
+    },
+    cancelMapUpdates () {
```
AlexZorkin: This method will cancel only half of the pendingMapUpdates because it is iterating forward while popping the last value in the array. Iterating in reverse would cancel all pending searches.
for (let i = this.pendingMapUpdates.length -1; i >= 0; i--) {
        const req = this.pendingMapUpdates.pop()
        req.cancel()
      }

stephenhillier: thanks, will fix and add test case

PR #1247, opened 2019-03-27
[1741] Allow Empty Form Fields on Edit

AlexZorkin:
- Added comparison against initial form values to check for real changes made
- Allowed empty string and form objects to be saved server side


app/backend/wells/stack.py
```diff
@@ -175,7 +175,7 @@ def _stack(self, records, well: Well) -> Well:
             for source_key, value in serializer.data.items():
                 # We only consider items with values, and keys that are in our target
                 # an exception is STAFF_EDIT submissions (we need to be able to accept empty values)
-                if value or value is False or value == 0:
+                if value or value is False or value == 0 or value == '' or value == {}:
```
stephenhillier: can we cover this with `value is not None` now?

Sybrand: I'm with Steve - at this point isn't not None equivalent?

AlexZorkin: yeah makes sense

PR #1255, opened 2019-03-29
[1720] Spatial Layers

Sybrand:
- Switched to using %s placeholders on query.
- Changed the schedule of exports to not overlap.
- Changed config to always pull image.
- Defining api definition for /gis/wells.
- Added in db comments as post_migrate step.
- Moved well table/column comments into model.
- Refactored export to use an iterator, allowing realtime streaming and simplifying code somewhat.
- Introduced base class code table model.
- Renamed some date fields to be consistent.
- Update fixtures to use expiry_date instead of expired_date.
- Added query options and realtime option
- Changed aquifer fixtures to have more continous numbering
- Added multipleOf based on db type


Sybrand:
[EPIC#Submission | Replication through DataBC - Spatial Layers - Final |3| !!!](https://trello.com/c/wTSEzu9j/1720-epicsubmission-replication-through-databc-spatial-layers-final-3)

PR #1257, opened 2019-03-29
[1752] BCeID Permissions

AlexZorkin:
- Separated submission permissions out of edit permissions
- Updated server views and client views with appropriate permission logic


app/backend/submissions/views.py
```diff
@@ -186,7 +186,7 @@ class SubmissionListAPIView(ListAPIView):
     get: returns a list of well activity submissions
     """
 
-    permission_classes = (WellsEditPermissions,)
+    permission_classes = (WellsSubmissionPermissions,)
```
stephenhillier: We should check with Lindsay but I think WellsEditPermissions is more appropriate for List and Get. I think we only want people with the "well submission" role to be able to submit reports, not list other reports.

Not sure if it's better to change the views to support different permission classes in a view class (for making new submissions vs listing them) or to change the permission class to only allow POST requests unless the user has edit rights.

AlexZorkin: I added a new role wells_submission_viewer which is required to access the get and list endpoints.

app/frontend/src/submissions/views/SubmissionsHome.vue
```diff
@@ -224,7 +224,7 @@ export default {
         if (this.isStaffEdit) {
           this.$noty.success('<div class="notifyText">Changes Saved!</div>', { killer: true })
         } else {
-          this.$noty.success('<div aria-label="Close" class="closeBtn">x</div><div class="notifyText">Well report submitted.</div>', { killer: true, timeout: false })
+          this.$noty.success('<div aria-label="Close" class="closeBtn">x</div><div class="notifyText">Well Report Submitted.</div>', { killer: true })
```
stephenhillier: It doesn't seem necessary or correct to capitalize all words in this sentence, was this a request from UX?  If so we can go with it.

PR #1261, opened 2019-04-01
Fix for Geom Migration

AlexZorkin:


app/backend/wells/migrations/0067_geom_updates.py
```diff
@@ -9,16 +9,18 @@
 def update_fields(apps, schema_editor):
     activitySubmission = apps.get_model('wells', 'activitysubmission')
     try:
+        activitySubmission._meta.get_field('longitude')
         activitySubmission._meta.get_field('latitude')
         schema_editor.execute("update activity_submission set geom = ST_SetSrid(ST_MakePoint(longitude, latitude), 4326);")
-    except FieldDoesNotExist:
+    except Exception:
```
Sybrand: If you don't specify that you're catching an Exception, you can catch an even broader number of issues.
```suggestion
    except:
```

app/backend/wells/migrations/0067_geom_updates.py
```diff
@@ -9,16 +9,18 @@
...
         logger.error('Field does not exist.')
 
     well = apps.get_model('wells', 'well')
     try:
+        well._meta.get_field('longitude')
         well._meta.get_field('latitude')
         schema_editor.execute("update well set geom = ST_SetSrid(ST_MakePoint(longitude, latitude), 4326);")
-    except FieldDoesNotExist:
+    except Exception:
```
Sybrand: I think we should just to an empty exception and catch everything
```suggestion
    except:
```

PR #1266, opened 2019-04-02
Aquifer Maps and Shapefiles

countable:


app/frontend/src/common/assets/js/leaflet-areaselect.js
```diff
@@ -0,0 +1,184 @@
+L.AreaSelect = L.Class.extend({
```
stephenhillier: need a blurb about the attribution of this file and the License (I think this is MIT licensed if I'm looking at the right repo)

app/frontend/src/aquifers/components/View.vue
```diff
@@ -13,7 +13,7 @@
 */
 
 <template>
-  <b-card class="container p-1">
+  <b-card class="container container-wide p-0 card-container pb-3" :class="{ 'p-4': editMode }">
```
stephenhillier: can you double check this works as intended (second `class` attribute)?

Second thing is that I think the container-wide class has been removed in our code base.  `container` now has an extra media query for 1080p screens with a max width of 1400px (App.vue).

stephenhillier:
Just need to add some attribution info for the file that was added, otherwise looks great

app/frontend/src/aquifers/components/View.vue
byshine: yes it does work as intended.

app/frontend/src/common/assets/js/leaflet-areaselect.js
byshine: where do you want the blurb? also what should it say?

stephenhillier:
```suggestion
/** 
 * leaflet-areaselect plugin by Jonatan Heyman
 * https://github.com/heyman/leaflet-areaselect
 * Source code made available under MIT license.
 */

L.AreaSelect = L.Class.extend({
```

stephenhillier:
Closing this PR since the other aquifers PR includes these commits.

PR #1269, opened 2019-04-02
[1730] add water quality characteristics to replication

stephenhillier:


app/database/scripts/wellsearch/wells_replication_stored_functions.sql
```diff
@@ -820,7 +820,40 @@ COMMENT ON FUNCTION migrate_drilling_methods () IS 'Load Well Drilling Method re
...
+--
+-- RETURNS
+--  None as this is a stored procedure
+--
+CREATE OR REPLACE FUNCTION migrate_water_quality_characteristics() RETURNS void as $$
+DECLARE
+  row_count integer;
+BEGIN
+  raise notice '...importing water_quality_characteristic';
+  INSERT INTO water_quality_characteristic(
```
Sybrand: water_quality_characteristic is a code table - so I don't think this will work. See line 109, app/backend/wells/models.py. 
There is a link table, called well_water_quality which I think is what you're intending to populate.

stephenhillier:
Table references have been fixed and I ran the replication with no issues (using the 10k well subset), and the water quality xref table was populated.

PR #1271, opened 2019-04-03
[1770] Replication of Pending, New, Rejected Well Records

AlexZorkin:


app/database/scripts/wellsearch/wells_replication_stored_functions.sql
```diff
@@ -334,6 +338,7 @@ BEGIN
      decommission_start_date            timestamp with time zone,
      decommission_end_date              timestamp with time zone,
      drilling_company_guid              uuid,
+     driller_name                       character varying(100),
```
stephenhillier: Is anything populating this column?  I think the closest match is `wells.wells_wells.crew_driller_name` but you'll have to double check that.

PR #1273, opened 2019-04-03
add redirect from /search to /

stephenhillier:


stephenhillier:
This PR adds https://gwells-dev-pr-1273.pathfinder.gov.bc.ca/gwells/search which redirects to the well search page.  This is the URL that the old wells site will redirect to

PR #1284, opened 2019-04-05
Style fixes for Advanced Search

cole:
Nicole pointed out that the date range on advanced search is extending past the border of the bootstrap card. This PR fixes that and also improves advanced search layout on small screens.

This will require some testing as the ApplicationAddEdit component was previously setting a global date input width, which applied to the entire site.


stephenhillier:
I tried the advance search tab and the ApplicationAddEdit component, looks good.

PR #1285, opened 2019-04-05
[1777] Fix for EMS ID display on well summary page

stephenhillier:
this PR changes the submission form to use `ems` instead of `ems_id` and it migrates data in `ems_id` over to the `ems` column.


app/backend/wells/migrations/0070_ems_id.py
```diff
@@ -0,0 +1,18 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wells', '0069_auto_20190405_2220'),
+    ]
+
+    operations = [
```
Sybrand: Isn't the column already gone at this point? It would have been renamed in migration 0069

app/backend/wells/models.py
```diff
@@ -1401,7 +1401,7 @@ class ActivitySubmission(AuditModel):
         WellYieldUnitCode, db_column='well_yield_unit_code', on_delete=models.CASCADE, blank=True, null=True)
     # want to be integer in future
     diameter = models.CharField(max_length=9, blank=True, null=True)
-    ems_id = models.CharField(max_length=30, blank=True, null=True)
+    ems = models.CharField(max_length=30, blank=True, null=True)
```
Sybrand: Why do we have to change it from ems_id to ems?

PR #1311, opened 2019-04-12
[1497] Well Licensing Information

AlexZorkin:
- Licensing information is now pulled from e-licensing and displayed on the well summary view.


app/backend/wells/views.py
```diff
@@ -632,3 +634,35 @@ def lithology_geojson(request):
             get_env_variable('S3_WELL_EXPORT_BUCKET'),
             'api/v1/gis/lithology.json')
         return HttpResponseRedirect(url)
+
+
+@api_view(['GET'])
+def well_licensing(request):
+    url = 'https://test.j200.gov.bc.ca/ws/' \
```
Sybrand: Can we make this a configuration setting please.

app/backend/wells/views.py
```diff
@@ -632,3 +634,35 @@ def lithology_geojson(request):
...
+          'RestGWellsInterface/wells/{}'.format(request.GET.get('well_tag_number'))
+    headers = {
+        'content_type': 'application/json',
+        'AuthUsername': get_env_variable('E_LICENSING_AUTH_USERNAME'),
+        'AuthPass': get_env_variable('E_LICENSING_AUTH_PASSWORD')
+    }
+    response = requests.get(url, headers=headers)
+    try:
+        licence = response.json()[-1]  # Use the latest licensing value
```
Sybrand: @stephenhillier I'm so tempted to update our wells licensing status here.... but that would be a crazy band-aid.... right?

AlexZorkin: Depends on where else it's visible in the app and if there would be syncing issues. If you can eliminate any out of sync scenarios then I'd say we could do the update.

app/backend/wells/views.py
```diff
@@ -634,3 +636,35 @@ def lithology_geojson(request):
             get_env_variable('S3_WELL_EXPORT_BUCKET'),
```
Sybrand: Please add env. variables to the openshift config: gwells/openshift/backend.dc.json

AlexZorkin: done

app/backend/wells/views.py
```diff
@@ -634,3 +636,35 @@ def lithology_geojson(request):
...
+    if licence:
+        licence_status = 'Licensed' if licence.get('authorization_status') == 'ACTIVE' else 'Unlicensed'
+        data = {
+            'status': licence_status,
+            'number': licence.get('authorization_number'),
+            'date': licence.get('authorization_status_date')
+        }
+    else:
+        data = {
+            'status': 'Unlicensed',
```
stephenhillier: We should expect the e-Licensing API to be unavailable sometimes, so we should only return "Unlicensed" status when get a successful response saying the well isn't licensed.  Maybe we could either respond with an error (503?) or use status "unavailable" (not sure we should be inventing licence statuses though)

stephenhillier: Currently the new `wells/licensing` endpoint returns 500 and the well detail page doesn't load:
```
TypeError: "t.licence is undefined"
    render WellDetail.vue:1
```

AlexZorkin: Updated backend to account for failed api calls. UX now reads "Service Unavailable" on failed api call to e-licensing.

AlexZorkin: Fixed front end undefined bug

PR #1315, opened 2019-04-16
Countable/licences (demo)

Sybrand:


byshine:
Added a bit more stuff in :)

PR #1316, opened 2019-04-17
[1786] Well Status Code Fix

AlexZorkin:


stephenhillier:
Testing by creating a new well with a construction report (status was set to "New"), edited something (status still "New"), then tested editing the status directly.  Everything worked

PR #1321, opened 2019-04-18
[1805] Performance increase & optimizations

Sybrand:
Performance increase & optimizations
    - Changed stack funtion to not use serializers as much.
    - Moved expensive to calculate maps into global constant maps.
    - Added lock to cached global variable.
    - Added tests.


Sybrand:
[EPIC#Submissions | Lithology from must be greater than 0 |2| !!](https://trello.com/c/gDvGxVFK/1805-epicsubmissions-lithology-from-must-be-greater-than-0-2)

PR #1330, opened 2019-04-30
[1824] Update meta-data as per DataBC requests

Sybrand:


Sybrand:
[EPIC#Other | Update meta data as per DataBC requests |1| !!!](https://trello.com/c/ehJhrAZG/1824-epicother-update-meta-data-as-per-databc-requests-1)

PR #1331, opened 2019-04-30
[1782] Well Edit History For Related Record Sets

AlexZorkin:


app/backend/wells/change_history.py
```diff
@@ -0,0 +1,152 @@
...
+    elif action == 'type_changes':
+        return 'Altered'
+    else:
+        return action
+
+
+# Ignores keys which we do not want to be flagged as diffable
+def keep_key(key):
+
+    return key != "root[0]['update_date']" and key != "root[0]['update_user']" \
```
Sybrand: How about a 
```
key not in (list, of, strings)
```
instead of and and and?

AlexZorkin: updated

PR #1341, opened 2019-05-02
Reset Connection by Peer Bug Fix

AlexZorkin:


app/backend/wells/views.py
```diff
@@ -650,6 +651,7 @@ def well_licensing(request):
         'AuthUsername': get_env_variable('E_LICENSING_AUTH_USERNAME'),
         'AuthPass': get_env_variable('E_LICENSING_AUTH_PASSWORD')
     }
+    time.sleep(0.01)  # hack to fix reset connection by peer error - server provider timeout issue
```
stephenhillier: yikes, not pretty, but it seems to be a popular hack for this issue

AlexZorkin: yeah i didn't find any other fix for this.

PR #1342, opened 2019-05-02
[1791] SSO Key Refresh

AlexZorkin:
This updates the SSO public key on app start. There could be additional work done to make this refresh automatically on a failed call but the method overrides required to automate this have a secret key conflict. JWT_PAYLOAD_HANDLER and JWT_RESPONSE_PAYLOAD_HANDLER are the methods to override if this is pursued further.


app/backend/gwells/settings/__init__.py
```diff
@@ -233,14 +235,26 @@
     }
 }
 
+
+try:
+    url = get_env_variable('SSO_AUTH_HOST') + '/realms/' + get_env_variable("SSO_REALM")
+    res = requests.get(url)
+    public_key = res.json()['public_key']
+    if public_key.__len__() <= 0:
```
stephenhillier:
```suggestion
    if len(public_key) <= 0:
```

PR #1349, opened 2019-05-07
[1809] Well Disinfected Updated to Code Table

AlexZorkin:


app/frontend/src/submissions/components/SubmissionForm/Completion.vue
```diff
@@ -123,11 +123,19 @@ Licensed under the Apache License, Version 2.0 (the "License");
...
-            <b-form-radio :value="true">Yes</b-form-radio>
-          </b-form-radio-group>
+        <b-form-group label="Well Disinfected Status" id="wellDisinfectedStatusInput">
+          <b-form-select
+            v-model="wellDisinfectedInput"
+            value-field="well_disinfected_code"
+            text-field="well_disinfected_code"
+            :errors="errors['well_disinfected_status']"
+            :loaded="fieldsLoaded['well_disinfected_status']">
+            <template slot="first">
```
Sybrand: I think we're pulling down the other codes from the API and setting them dynamically instead of hard coding them? Technically, if we do that, the API shouldn't return the "Unknown" state, since it's been expired, so I'm not sure how we would deal with that on the front end other than forcing the user to pick Not Disinfected as a default?

PR #1352, opened 2019-05-08
[1558] Export Well Search Results

cole:
Only contains CSV export for now (PR submitted for initial UX review).


cole:
[EPIC#AdvancedSearch | Well Results - Export Results |5| !!! [BV: 22]](https://trello.com/c/kigqn9OC/1558-epicadvancedsearch-well-results-export-results-5-bv-22)

cole:
Updated with Excel export and ready for review.

Note that although this is fast enough to not timeout on dev now with the default column set, it will likely time out on stage, or if someone selects all columns. It seems like there is a ~5 second timeout coming from something (haproxy?) that will probably need to be extended for these exports.

cole:
@Sybrand @stephenhillier I think this is ready for merging now.

PR #1353, opened 2019-05-08
Well Licensing will default to DB Value if E-Licensing API Fails

AlexZorkin:


app/backend/wells/views.py
```diff
@@ -638,34 +638,38 @@ def lithology_geojson(request):
...
-        except:
-            data = {
-                'status': 'Unlicensed',
-                'number': 'N/A',
-                'date': ''
-            }
-    else:
+
+    try:
+        response = requests.get(url, headers=headers)
```
Sybrand: Is the server provider timeout fix no longer required?
```
time.sleep(0.01)  # hack to fix reset connection by peer error - server provider timeout issue
```

AlexZorkin: I don't believe so because the real issue is the firewall settings. Once the firewall has been opened up properly then we'll see if the reset connection bug is still an issue.

app/backend/wells/views.py
```diff
@@ -638,34 +638,38 @@ def lithology_geojson(request):
...
+            except:
+                pass
+    except:
+        pass
+
+    if not api_success:
         data = {
-            'status': 'Service Unavailable',
-            'number': 'N/A',
+            'status': well.licenced_status.description,
```
Sybrand: Can we move the well lookup into the fail section, we only need to load it if the api fails after all.

PR #1354, opened 2019-05-08
[1828] update_user and create_user on related records

Sybrand:
- Added testing for lithology
- Changed how serialization works for related records


app/backend/submissions/tests.py
```diff
@@ -322,7 +379,7 @@ def test_no_city_on_legacy(self):
             'well': well.well_tag_number
         }
 
-        # NOTE: this test previously asserted that a 
+        # NOTE: this test previously asserted that a
```
stephenhillier:
```suggestion
```

I think I accidently left this half-finished comment (I moved it to the docstring but left a few words down here)

PR #1358, opened 2019-05-10
[1825][1826] Ground Elevation Validation

AlexZorkin:
1. Added ‘Unknown’ as a value to the ground elevation method code table
2. Replaced dummy ground_elevation 0 values in the database with null values
3. Added validation to make Ground Elevation and Method for Determining Ground Elevation fields mutually required. Exception is if ‘Unknown’ is selected as the method, this allows saving with an empty ground_elevation field.
4. Checking for empty string on ground_elevation and swapping it to a null value before save.
5. Ground Elevation input field only takes numbers now

** Waiting for Stacker Changes to take effect to allow saving of null values **
** Clearing ground_elevation won't save until this is altered **


app/backend/submissions/views.py
```diff
@@ -300,6 +300,13 @@ class SubmissionStaffEditAPIView(SubmissionBase):
     permission_classes = (WellsEditPermissions,)
     queryset = ActivitySubmission.objects.all()
 
+    def post(self, request, *args, **kwargs):
+        # ground_elevation is a decimal so we swap empty string with null value
+        if 'ground_elevation' in request.data:
+            if request.data['ground_elevation'] == '':
```
Sybrand:
```suggestion
            if request.data['ground_elevation'] === '':
```

stephenhillier:
@AlexZorkin  could you update this branch?  I pushed a fix to the release branch that should solve the deployment error.

PR #1360, opened 2019-05-14
Well Disinfect View Logic

AlexZorkin:


app/frontend/src/submissions/components/SubmissionForm/Completion.vue
```diff
@@ -192,9 +192,11 @@ export default {
   },
   methods: {
     disinfected_codes () {
-      this.codes.well_disinfected_codes.forEach((code) => {
+      this.codes.well_disinfected_codes.forEach((code, index, object) => {
```
Sybrand: @AlexZorkin @stephenhillier I think we talked about this - I don't like it - since it should really be applied on the API, not the front-end - but if this is the fastest way to resolve it - I'm OK with it.

PR #1361, opened 2019-05-14
[1766] Class of Well Validation and Data Cleanup

AlexZorkin:


app/backend/wells/migrations/0084_update_intended_water_use.py
```diff
@@ -0,0 +1,44 @@
+from django.db import migrations
+
+
+# This can be deleted when doing next squash of migrations because it's a one time update
+def update_intended_water_use_code(apps, schema_editor):
+    intended_water_use = apps.get_model('wells', 'IntendedWaterUseCode')
+    well = apps.get_model('wells', 'well')
+
+    not_applicable = intended_water_use.objects.get_or_create(
```
Sybrand: @AlexZorkin do we need this in a fixture? Just thinking, once this gets squashed, we'll probably need to load it up in our environments?

AlexZorkin: yup i missed that ty, ill add it in.

PR #1371, opened 2019-05-16
[1776] Screen Data Cleanup

AlexZorkin:


app/backend/wells/management/commands/export.py
```diff
@@ -238,7 +238,7 @@ def generate_files(self, zip_filename, spreadsheet_filename):
         ########
         # SCREEN
         ########
-        screen_sql = ("""select screen.well_tag_number, screen_from, screen_to, internal_diameter,
+        screen_sql = ("""select screen.well_tag_number, screen_from, screen_to, diameter,
```
Sybrand: Shouldn't this be: screen_diameter?
```suggestion
        screen_sql = ("""select screen.well_tag_number, screen_from, screen_to, screen_diameter,
```

AlexZorkin: yes good catch ty

PR #1372, opened 2019-05-16
[1803] Fixed Pump Installer Registration #

AlexZorkin:
also added in reg numbers to fixtures for testing.


app/frontend/src/submissions/components/SubmissionForm/PersonResponsible.vue
```diff
@@ -239,7 +239,7 @@ export default {
     personNameReg (option) {
       if (option && option.registrations) {
         const drillReg = option.registrations.find((item) => {
-          return item.registries_activity === 'DRILL'
+          return item.registries_activity === 'DRILL' || item.registries_activity === 'PUMP'
```
Sybrand: I'm trying to rack my brain why this even matters? Why filter at all? If there's a reg number, why not just use it? Why care if ti's a driller or pump installer?

PR #1388, opened 2019-05-31
fix serializer bug discovered by Steve H

countable:


stephenhillier:
Going to put this on the back burner for now as this bug was solved another way and the proposed changes here will need corresponding test updates, thanks for putting forward a fix however!  #1385

PR #1402, opened 2019-06-10
rolling deployment

stephenhillier:
this PR changes the deployment strategy to 'Rolling', which means there won't be any downtime while deploying new versions. DB migrations are handled in the Jenkinsfile before deploying, so all migrations need to be compatible with the existing code.


stephenhillier:
[EPIC#Pipeline | Rolling deployment |2| !!](https://trello.com/c/5We5pIC4/1897-epicpipeline-rolling-deployment-2)

PR #1412, opened 2019-06-19
[1847] Acquisition code default set to 'H'

AlexZorkin:


AlexZorkin:
Business are wants to change the code value, so i just need to update that.

PR #1439, opened 2019-10-10
Steve/db comments2

stephenhillier:


app/backend/aquifers/models.py
```diff
@@ -134,12 +134,7 @@ class AquiferSubtype(CodeTableModel):
...
-                    ' proximity and likelihood of hydraulic connection to streams. The code value is a'
-                    ' combination of an aquifer type represented by a number and an optional letter'
-                    ' representing a more specific aquifer sub-type. E.g. 1a, 2, 6a.'))
+        db_comment=("Categorizes an aquifer based on how it was formed geologically (depositional description). Understanding of how aquifers were formed governs important attributes such as their productivity, vulnerability to contamination as well as proximity and likelihood of hydraulic connection to streams. The code value is a combination of an aquifer type represented by a number and an optional letter representing a more specific aquifer sub-type.   There are six major aquifer types, some with multiple subtypes.  E.g. aquifer sub-type code 6b is comprised of the aquifer type number (6: Crystalline bedrock aquifers) and subtype letter (b) specifically described as: Fractured crystalline (igneous intrusive or metamorphic, meta-sedimentary, meta-volcanic, volcanic) rock aquifers. Code values range from 1a to 6b."))
```
dainetrinidad: The formatting for the new code looks different from the old code. Double quotes vs single and it also ignores the line limit

stephenhillier: I've been looking but so far I haven't found any tools to break apart the long strings automatically.  It looks like the main autoformatters leave long strings intact (there's about 250 of the db comment strings so a more automated solution would be great).

PR #1440, opened 2019-10-16
Fixed spelling error for filter pack

dainetrinidad:
Looks like the data was imported previously with a spelling error. I simply reformatted the JSON file to make it easier to read, and created a new migration to unload & then reload the fixture.

We need to run it on staging and prod for it to take effect


stephenhillier:
looks good, but could you rebase off the release branch?  It looks like this was branched from master, so it has extra merge commits. Not sure if that affects anything other than just being a bit messy.

dainetrinidad:
Moved this change into https://github.com/bcgov/gwells/pull/1443

PR #1443, opened 2019-10-21
Reloading filter packs using migration to update fixture

dainetrinidad:
**Fixed spelling error for filter pack**

Looks like the data was imported previously with a spelling error. I simply reformatted the JSON file to make it easier to read, and created a new migration to unload & then reload the fixture.

We need to run it on staging and prod for it to take effect


dainetrinidad:
@stephenhillier I don't think it does a delete cascade, it uses the custom code to load & unload fixtures which just deal with a specific model. Since the PK is in the fixture json, and doesn't change, it shouldn't affect any dependent tables.

```python
# gwells/codes.py
    def unload_fixture(self, apps, schema_editor):
        for item in self.fixture:
            app, model = item.get('model').split('.', 1)
            model = apps.get_model(app, model)
            # The fixture in question may have been deleted for other reasons, so we don't force a
            # delete. If the record exists, then delete it.
            item = model.objects.filter(pk=item.get('pk')).first()
            if item:
                item.delete()
```

PR #1447, opened 2019-10-30
OBS Well Number and Status added to Databc api call and swagger docs

AlexZorkin:


stephenhillier:
The observation well status is a code table reference
`well.obs_well_status_code`
They may want the description from the code table (possibly?)

PR #1453, opened 2019-11-21
Fixes to Cascading Behaviour in Models

AlexZorkin:
Updated majority of on_delete=models.CASCADE to on_delete=models.PROTECT
This cascading behaviour was causing errant deletions of wells and activity_submissions when child objects were deleted.


seamus-oconnor:
@stephenhillier or @AlexZorkin with the package-lock.json removed from this PR can it be merged now?

PR #1457, opened 2019-11-28
Categorizes observation wells into Active/Inactive

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-540

Screen Shot 2019-11-27 at 3 45 36 PM


seamus-oconnor:
Blocked until fixtures are updated with an observation well that is associated with an aquifer.

seamus-oconnor:
@stephenhillier it finally passes!

stephenhillier:
I don't think this was part of this PR (probably the other v1/v2 PR?) but can we make the default version remain as V1?  That way our docs will still say v1 and not update to v2 just yet.

seamus-oconnor:
> I don't think this was part of this PR (probably the other v1/v2 PR?) but can we make the default backend API version remain as V1? That way our docs will still say v1 and not update to v2 just yet.

Yes. The API now defaults to v1. The front-end is still using v2.

app/backend/aquifers/views.py
```diff
@@ -161,7 +166,9 @@ def get_serializer_class(self):
         if self.request.method == 'GET':
             return serializers.AquiferSerializer
         else:
-            return serializers.AquiferDetailSerializer
+            if self.request.version == 'v1':
+                return serializers.AquiferDetailSerializerV1
+            return serializers_v2.AquiferDetailSerializerV2
```
dainetrinidad: This looks fine, but is it possible to organize this better? I know there's currently no `v1/` folder, but how complex would it be to put all the new code in a new folder called `v2/`? So all new serializers, models, etc would reside there. Thoughts?

stephenhillier: Since it's just the serializers file that needs a new v2 file I don't think that's necessary yet, but we should definitely be prepared to re-organize things if the v2 serializers get messy.  This might start happening if we need different view methods for v2/v1.

PR #1460, opened 2019-11-29
Updates the API to support both v1 and v2 in the path

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-688


seamus-oconnor:
@stephenhillier tests are failing because of the new `request` arg [required by versioning](https://www.django-rest-framework.org/api-guide/versioning/#reversing-urls-for-versioned-apis).

seamus-oconnor:
@stephenhillier tests are now failing because of three Postman test failures:

1.  JSONError
  Unexpected token '<' at 1:1
  \<!DOCTYPE html\>
  ^
  at test-script
  inside "Activity Submissions / Staff Edit / Create aquifer for well"

2.  AssertionError  Status code is 201
  expected 400 to equal 201
  at assertion:0 in test-script
  inside "Activity Submissions / Staff Edit / Staff Edit - Associate well with aquifer"

3.  AssertionError  Expect storativity to be 1
  expected NaN to equal 1
  at assertion:2 in test-script
  inside "Activity Submissions / Staff Edit / Staff Edit - Associate well

seamus-oconnor:
@stephenhillier the build is passing now! :tada:

app/backend/aquifers/urls.py
```diff
@@ -18,102 +18,103 @@
 from aquifers import views
 
 CACHE_TTL = 60*15
+api_version_url_re = '(?P<version>(v1|v2))'
 
 urlpatterns = [
-    url(r'^api/v1/aquifers$',
+    url(rf'^api/' + api_version_url_re + r'/aquifers$',
```
dainetrinidad: Minor inconsistency with the other lines of code

app/backend/wells/tests/test_documents.py
```diff
@@ -13,13 +13,18 @@
...
+
 from rest_framework import status
 from rest_framework.test import APITestCase
+from rest_framework.reverse import reverse
 
+from gwells.settings import REST_FRAMEWORK
 from gwells.documents import MinioClient
 
 
+API_VERSION = REST_FRAMEWORK['DEFAULT_VERSION']
```
dainetrinidad: I think we might have to test agains both v1 & v1 and not just the default/current version

seamus-oconnor: Since the v2 APIs are identical to v1 (outside of the upcoming minor change to /aquifers/<num>) I think we should only be testing the latest version, except in cases where the previous version API is different then the current one.

Otherwise we will have to double (then triple and so on) the API tests but for very little value of testing something uniquely different for each API.

app/backend/aquifers/urls.py
seamus-oconnor: Good catch - this syntax though is changing in the next commit after a discussion with Steve.

app/backend/wells/tests/test_documents.py
stephenhillier: The postman tests might be a good place to test against both v1 & v2, I can help with that (probably tomorrow)

seamus-oconnor: We could verify that the output of a /v2/ API call is the same as a /v1/ call. That might be an easy way (so we don't duplicate tests) of verifying that the API hasn't changed in ways that are unexpected.

PR #1464, opened 2019-11-29
Scrolls to the top of the aquifer's results

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-529

### DIFF
--- a/app/frontend/src/aquifers/components/Search.vue
+++ b/app/frontend/src/aquifers/components/Search.vue
@@ -313,8 +313,14 @@ export default {
             aquifer.gs = geomCache[aquifer.id]
           })
           this.response = response.data
-          this.scrollToTableTop()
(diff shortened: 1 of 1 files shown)


app/frontend/src/aquifers/components/Search.vue
```diff
@@ -313,8 +313,14 @@ export default {
             aquifer.gs = geomCache[aquifer.id]
           })
           this.response = response.data
-          this.scrollToTableTop()
           this.loading = false
+          if (response.data.results.length > 0) {
```
dainetrinidad: Seems like this is assigned to `this.aquifers_search_results` above, might be cleaner to change it to
```
if (this.aquifers_search_results.length > 0) {
```

PR #1465, opened 2019-12-02
Improves print CSS for aquifer summary page

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-527

Screenshot from 2019-12-02 15-56-00

Example print PDF:
[Groundwater Wells and Aquifers - Province of British Columbia.pdf](https://github.com/bcgov/gwells/files/3913829/Groundwater.Wells.and.Aquifers.-.Province.of.British.Columbia.pdf)

### DIFF
--- a/app/frontend/package-lock.json
+++ b/app/frontend/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "^7.0.0"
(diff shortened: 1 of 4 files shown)


seamus-oconnor:
@stephenhillier I added a leaflet plugin to load tiles that are outside the map's container. This was required because the map is expanded when printing leading to missing tiles in the PDF. However this caused the package-lock.json to be updated when I ran `docker-compose exec frontend npm install --save leaflet-edgebuffer`.

stephenhillier:
Looks good so far.  I think this will only print in Chrome but we have the same problem with the Well detail page (only prints 1 page on other browsers like Firefox)

seamus-oconnor:
@stephenhillier it looks like FF (and maybe other browsers) can't handle flexbox spanning multiple pages. I have added some specific rules for only print to change flex → block and it now prints all the pages in FF (on my computer at least).

PR #1474, opened 2019-12-04
CSS updates to knowledge indicators

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-538

Screen Shot 2019-12-04 at 2 15 48 PM

### DIFF
--- a/app/frontend/src/aquifers/components/View.vue
+++ b/app/frontend/src/aquifers/components/View.vue
@@ -198,7 +198,7 @@
             <div class="aquifer-information-list-divider"></div>
             <li :key="section.id" v-for="section in aquifer_resource_sections">
               <div class="observational-wells" v-if="section.key === 'obs-wells'">
-                <dt>Observation wells</dt>
+                <dt class="text-right">Observation wells</dt>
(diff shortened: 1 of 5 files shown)


app/frontend/src/aquifers/components/View.vue
```diff
@@ -307,18 +314,23 @@ a {
 }
 
 .aquifer-information-list dt {
+  padding-right: 2rem;
   font-weight: bold;
 }
 
 .aquifer-information-list dd {
-  padding-left: 1rem;
+  padding-left: 2rem;
```
dainetrinidad: Vuetify also has spacing helpers you can use (which seems to be heavily used up there) but could potentially get rid of all these css with margin-bottom: 0

seamus-oconnor: I checked the Bootstrap Vue [text spacing documentation](https://getbootstrap.com/docs/4.3/utilities/spacing/) and it shows that there is no way to get a padding of `2rem`. A class of `pr-4` = `1.5rem` and `pr-5` = `3rem`.

PR #1478, opened 2019-12-06
Use long form for volumes definition so non-macOS OSes can read the docker-compose.yml file

seamus-oconnor:
Strangely docker compose on my Linux machine doesn't understand the shorthand volume syntax. I needed to change it to the long form version in order for `docker-compose up` to work.


seamus-oconnor:
The docker-compose.yml file is only used by developers for their local docker environment. Openshift uses a different set of configs to build images.

PR #1481, opened 2019-12-11
540 fix obs wells not displaying inactive wells bug

seamus-oconnor:
Verifying [WATER-540](https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-540) with staging data I saw that the algorithm I was using wasn't working for aquifers with only inactive observation well data (e.g. aquifer #612). I have fixed the algorithm so the display of active and inactive wells shows up for any obs wells data.

image

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+API_TARGET=docker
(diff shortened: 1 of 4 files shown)


seamus-oconnor:
This task is waiting for PR #1480 to be merged first as I had to use the API_TARGET env var feature in order to fix this observation well bugs.

PR #1485, opened 2019-12-12
Aquifer search improvements (all and any radios)

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-530

Adds ability to search by "any" and "all" matching sections (search check boxes)

image

### DIFF
--- a/app/backend/aquifers/views.py
+++ b/app/backend/aquifers/views.py
@@ -117,7 +117,7 @@ def _aquifer_qs(query):
     search = query.get('search')
 
     match_any = True
-    if request.version == 'v2':
+    if self.request.version == 'v2':
         # V2 changes to `and`-ing the filters by default unless "match_any" is explicitly set to 'true'
         match_any = query.get('match_any') == 'true'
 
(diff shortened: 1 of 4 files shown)


app/backend/aquifers/views.py
```diff
@@ -112,20 +112,34 @@ def _aquifer_qs(query):
...
-    # truthy check - ignore missing and emptystring.
+    # ignore missing and empty string for resources__section__code qs param
     if resources__section__code:
-        qs = qs.filter(
-            resources__section__code__in=resources__section__code.split(','))
+        for code in resources__section__code.split(','):
+            filters.append(Q(resources__section__code=code))
 
-    if search:  # truthy check - ignore missing and emptystring.
+    if match_any:
```
stephenhillier: might need a check to see if there are any filters selected? (if filters is still an empty list you may get an error calling `pop()`)

seamus-oconnor: Good catch

stephenhillier:
Looks good overall, with just the one suggestion that we fix the match any behavior for v2 only and leave v1 behaving relatively the same as before.

PR #1486, opened 2019-12-12
add extra data to aquifer/wells test fixtures

stephenhillier:


seamus-oconnor:
I pulled down the latest changes and I still don't see licensed wells showing up for aquifer 1901 in the UI or the API request: http://localhost:8080/api/aquifers/1901

seamus-oconnor:
While you are changing this fixture data 1900's geom is out in the Pacific ocean somewhere. It would be helpful for this aquifer to have geometry inside of BC somewhere.

PR #1489, opened 2019-12-13
Well Lithology Endpoint added to API

AlexZorkin:


app/backend/wells/urls.py
```diff
@@ -45,6 +45,11 @@
     url(api_path_prefix() + r'/wells/screens$',
         never_cache(views.WellScreens.as_view()), name='well-screens'),
 
+    # Well lithology search
+    # returns information about well lithology for a range of wells provided in ?wells=123,124,125 etc.
+    url(api_path_prefix() + r'/wells/lithology$',
```
seamus-oconnor: Should this API be limited to v2 only?

PR #1494, opened 2019-12-17
Adds Ground/Surface water interaction section

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-685

image

image


app/backend/aquifers/migrations/0026_auto_20191216_2135.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+    VALUES (
+        'G',
+        'DATALOAD_USER',
+        '2020-01-01T08:00:00Z',
+        'DATALOAD_USER',
+        '2020-01-01T08:00:00Z',
+        '1970-01-01T08:00:00Z',
+        '9999-12-31T23:59:59Z',
+        'Groundwater Surface Water Interactions',
```
stephenhillier: I think this should be Groundwater surface water interactions, since it's not a title.

app/frontend/src/aquifers/components/View.vue
```diff
@@ -482,6 +482,7 @@ export default {
         { code: 'W', name: 'Water budget' },
         { key: 'water-quality', name: 'Water quality information' },
         { key: 'aquifer-connected', name: 'Hydraulically connected (screening level)' },
+        { code: 'G', name: 'Groundwater Surface Water Interactions' },
```
stephenhillier: same comment here (capitalization)

PR #1495, opened 2019-12-17
Steve/well api v2

AlexZorkin:


app/backend/wells/views.py
```diff
@@ -361,7 +367,37 @@ def get(self, request, **kwargs):
         if count == 0:
             raise NotFound("No well records could be found.")
 
-        return super().get(request)
```
seamus-oconnor: This looks like it will change the v1 response - should this be isolated to just v2?

stephenhillier: I'll move this code into the V2 view once that PR is finished (however I did try to prevent the response from changing, unless requesting geojson)

PR #1496, opened 2019-12-17
Improve Groundwater licences in legend

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-708

_thumb_34168


app/frontend/src/aquifers/layers.js
```diff
@@ -3,7 +3,8 @@ import L from 'leaflet'
 import ArtesianLegend from '../common/assets/images/artesian.png'
 import CadastralLegend from '../common/assets/images/cadastral.png'
 import EcocatWaterLegend from '../common/assets/images/ecocat-water.png'
-import GWaterLicenceLegend from '../common/assets/images/gwater-licence.png'
+import GWaterLicenceActiveLegend from '../common/assets/images/gwater-licence-active.svg'
+import GWaterLicenceInactiveLegend from '../common/assets/images/gwater-licence-inactive.svg'
 import OWellsInactiveLegend from '../common/assets/images/owells-inactive.png'
 import OWellsActiveLegend from '../common/assets/images/owells-active.png'
```
dainetrinidad: It's usually more ideal to spell it out, unless this is how it is all across the gwells app?
GWaterLicence vs GroundWaterLicence
OWells vs ObservationWells (should it be ObservationWell since the others are not plural?)

app/frontend/src/aquifers/legend.js
```diff
@@ -0,0 +1,24 @@
+export function buildLegendHTML (items) {
```
dainetrinidad: What's the reason behind doing it this way vs creating a vue template?

app/frontend/src/aquifers/layers.js
seamus-oconnor: I was copying the structure of what was there. Personally it isn't a `class` so I wouldn't use an initial capital on the first letter either. There are a lot of short names used through-out GWELLS. Would you like to me to fix this as I go? Spelling things out and fixing weird capitalization?

app/frontend/src/aquifers/legend.js
seamus-oconnor: Yes. This pattern follows what was done before (and it was duplicated between AquiferMap.vue and SingleAquiferMap.vue). I attempted to use a Vue component for the legend but after looking for examples online and using the leaflet2-vue project as a guide there was no "good" way of doing it. It doesn't seem to me that Vue is supposed to be integrated into non-Vue DOM - at least not officially. This function is at least an improvement on what was there as it removes duplication and produces cleaner HTML.

app/frontend/src/aquifers/layers.js
dainetrinidad: I think that would be beneficial to fix them as you go

dainetrinidad:
Opening a technical debt task for some of the discussions around leaflet in gwells
https://apps.nrs.gov.bc.ca/int/jira/browse/GWELLS-38

PR #1500, opened 2019-12-19
Pins leaflet-gesture-handling npm module

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-749

Fixes a bug where you can no longer drag when on a non-vector layer part of the map if you initially click down on a vector layer.

### DIFF
--- a/app/frontend/package.json
+++ b/app/frontend/package.json
@@ -24,7 +24,7 @@
     "leaflet-edgebuffer": "^1.0.6",
     "leaflet-fullscreen": "^1.0.2",
     "leaflet-geosearch": "^2.7.0",
-    "leaflet-gesture-handling": "^1.1.8",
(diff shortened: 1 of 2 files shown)


app/frontend/package.json
```diff
@@ -24,7 +24,7 @@
     "leaflet-edgebuffer": "^1.0.6",
     "leaflet-fullscreen": "^1.0.2",
     "leaflet-geosearch": "^2.7.0",
-    "leaflet-gesture-handling": "^1.1.8",
+    "leaflet-gesture-handling": "mrleblanc101/Leaflet.GestureHandling#c78f7414c8615a56d4eab54f4be5b9b34149afec",
```
dainetrinidad: Would this create a technical debt in the future given it's not the official version?

seamus-oconnor: It is a possibility. But the main project project has been 1.1.8 for over a year. There is no signs in the issue tracker that the project owner is updating it anymore either. The bug is severe enough that we should either remove the plugin completely or live with this forked version IMO.

PR #1503, opened 2019-12-23
Shows searched aquifers on map

seamus-oconnor:
Aquifers are orange by default. Searched aquifers are shown in purple. The selected aquifer (after clicking the table row) is green. Clicked aquifers show a popup with a link to that aquifer.

The search button has been changed to by default search in all of BC. When the map is zoomed into a sub-section of BC the label changes to "Search In Map Area". The current visible area is then used to narrow the search results. Clicking the "Reset" button will zoom the map back out to the province allowing for a province wide search.

image


seamus-oconnor:
This PR builds off of #1501. Don't review or merge it until #1501 is merged into release.

app/backend/aquifers/filters.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+    def filter_queryset(self, request, queryset, _view):
+        sw_long = request.query_params.get('sw_long')
+        sw_lat = request.query_params.get('sw_lat')
+        ne_long = request.query_params.get('ne_long')
+        ne_lat = request.query_params.get('ne_lat')
+
+        if sw_long and sw_lat and ne_long and ne_lat:
+            ct = CoordTransform(SpatialReference('NAD83'), SpatialReference('EPSG:4326'))
```
seamus-oconnor: @stephenhillier is this the right kind of transform to use here? It seems to work but I am assuming that the query string parameters are in 4326 and it looks like the geom data in the DB for aquifers is NAD83. It works for the local aquifer fixtures but I don't know for sure if this is doing the right thing. Some transform is needed because without it every bounding box query string I tried returns no aquifers.

stephenhillier: The aquifer layer geometry uses SRID 3005 (BC Albers, which uses metres as a base unit) while the coordinates in the query (`sw_long`, `sw_lat` etc) are SRID 4326 (and are in degrees). By specifying that the bounding box is SRID 4326, GeoDjango can transform the bounding box to 3005 to do the intersect.

With that in mind, just setting the `srid` property on the polygon you made from the bounding box (i.e. `bbox.srid = 4326`, without the `bbox.transform(ct)` statement) seems more like what you need to do here. However... there are people more knowledgeable than me who can clarify this for us next week.

seamus-oconnor: Yes that change worked well - thx!

PR #1505, opened 2019-12-30
Add lat/long and geometry filter to screens (drawdown) endpoint

stephenhillier:


app/backend/wells/filters.py
```diff
@@ -102,6 +104,30 @@ def filter_queryset(self, request, queryset, view):
...
+        point = request.query_params.get('point', None)
+        radius = request.query_params.get('radius', None)
+        srid = request.query_params.get('srid', 4326)
+
+        if point and radius:
+            try:
+                shape = GEOSGeometry(point, srid=int(srid))
+                assert shape.geom_type == 'Point'
+            except (ValueError, AssertionError, GEOSException):
+                pass
```
dainetrinidad: Would it be beneficial to log the error instead of `pass`?

seamus-oconnor: It might not be since it is coming in from the URL and therefore any junk from the client could pollute the logs. This code is very similar to what was already in place by [BoundingBoxFilterBackend](https://github.com/bcgov/gwells/blob/122d318a73cda2a16a0c404b293d534a31c8eb0b/app/backend/wells/filters.py#L63-L82) for wells. If we change this logging / not behaviour we should change that one as well.

dainetrinidad: I see, so it doesn't matter that much if the AssertionError occurs then? Would you still need the `assert` if that's the case?

seamus-oconnor: @stephenhillier might be able to answer that. It looks like it is making sure that the `"point"` QS param is indeed an actual point geometry via the `assert`.

stephenhillier: That was the goal, although it might be more clear while reading this code if it was just another if statement.

Upon reflection the error handling could be improved, maybe at the least a debug log message?

PR #1509, opened 2020-01-06
Improves back button behaviour

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-691

Well and aquifer search results, details and map position are stored in Vuex for quick re-display on browser back.


seamus-oconnor:
This PR is built off of #1503 and it needs to merge first.

app/backend/wells/views.py
```diff
@@ -764,7 +764,7 @@ def lithology_geojson(request, **kwargs):
 @api_view(['GET'])
 def well_licensing(request, **kwargs):
     tag = request.GET.get('well_tag_number')
-    url = get_env_variable('E_LICENSING_URL') + '{}'.format(tag)
+    eLicensingUrl = get_env_variable('E_LICENSING_URL')
```
dainetrinidad:
```suggestion
    e_licensing_url = get_env_variable('E_LICENSING_URL')
```

app/frontend/src/wells/views/WellSearch.vue
```diff
@@ -228,8 +253,55 @@ export default {
...
+
+      this.$store.commit(SET_SEARCH_PARAMS, query)
+    },
+    updateQueryParams () {
+      const tabHash = (this.tabIndex === 1) ? 'advanced' : null
 
-      this.$store.commit(SET_SEARCH_PARAMS, { ...query })
+      this.$router.replace({ query: this.buildQueryParams(), hash: tabHash })
+    },
+    buildQueryParams () {
```
dainetrinidad: Would it make sense to have a unit test for this method?

app/frontend/src/wells/views/WellSearch.vue
```diff
@@ -289,38 +338,36 @@ export default {
...
           })
         }
-      }).catch((e) => {
-        console.error(e)
       })
+    },
+    handleRouteChange () {
+      const query = this.$route.query
+      const emptyQuery = Object.keys(query).length === 0
+      const storeStateAsQS = this.buildQueryParams()
```
dainetrinidad: It seems like the state is being handled through this method, so it would be nice if it's possible to have a unit test for it

app/frontend/src/wells/views/WellSearch.vue
seamus-oconnor: I have added a test that loads the component and triggers a `handleMapSearch` method on that component. That will force an update of the URL's query params calling `buildQueryParams`

app/frontend/src/wells/views/WellSearch.vue
seamus-oconnor: I have added a test that loads the component and triggers a `handleMapSearch` method on that component. That will force an update of the URL's query params calling `buildQueryParams`

app/frontend/src/aquifers/components/AquiferMap.vue
```diff
@@ -6,6 +6,7 @@
 import L from 'leaflet'
 import 'leaflet-gesture-handling'
 import { tiledMapLayer } from 'esri-leaflet'
+import { uniq } from 'lodash'
```
stephenhillier: I'm guessing lodash was already installed but I think I would prefer to install just `lodash.uniq` instead and not load the complete library.

seamus-oconnor: I see that we are using specific lodash stuff in other places but this isn't the only place that has imported `from 'lodash'`. I feel like we will end up with a lot of package.json pollution by `npm -i lodash.*` libs as needed.

We should discuss and maybe we can find a nicer way of slimming down the lodash dep size using a Webpack/Babel plugin to do the hard work for us.

seamus-oconnor: I installed lodash-es which w/ Webpack 4's tree shaking reduced the lodash size in the bundle from 188kb to 44kb. I also aliased moment to moment/src/moment which includes only the EN locale reducing it's size from 193kb to 59kb.

PR #1517, opened 2020-01-11
Derives UTM easting / northing from geom

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-700

and

https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-773


app/backend/wells/models.py
```diff
@@ -1248,6 +1252,18 @@ def longitude(self):
...
 
+@receiver(pre_save, sender=Well)
+def update_utm(sender, instance, **kwargs):
+    if instance.geom and (-180 < instance.geom.x < 180): # only update utm when geom is valid
+        utm_zone = math.floor((instance.geom.x + 180) / 6) + 1
+        coord_transform = CoordTransform(SpatialReference(4326), SpatialReference(32600 + utm_zone))
+        utm_point = instance.geom.transform(coord_transform, clone=True)
+
+        instance.utm_zone_code = utm_zone
+        instance.utm_easting = round(utm_point.x)
```
dainetrinidad: Might be a good idea to add some comments in this area. I'm a little concerned about the loss of precision, which might come up as an issue in the future (or not)

seamus-oconnor: Because easting and northing is only precise to 1m there is loss no matter what. The DB field is an integer type as well. I can add comments to this function. :-D

dainetrinidad: We can just add what you just said and consider it done

PR #1518, opened 2020-01-13
Fixes UTM input bug

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-482

Replaces Coords.vue input locking & watches with on-input coords conversion logic

### DIFF
--- a/app/frontend/src/submissions/components/SubmissionForm/Coords.vue
+++ b/app/frontend/src/submissions/components/SubmissionForm/Coords.vue
@@ -274,7 +274,7 @@ export default {
       utm: {
         easting: null,
         northing: null,
-        zone: ''
+        zone: null
       },
       latitudeDMSValidation: false,
       coordinateLookup: new Map(),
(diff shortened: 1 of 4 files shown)


app/frontend/src/submissions/components/SubmissionForm/Coords.vue
```diff
@@ -508,8 +340,50 @@ export default {
         zone: null
       }
     },
+    handleDegreesChange () {
```
dainetrinidad: Would be nice if there were unit tests for these new methods

seamus-oconnor: They are not new - just changed from the watch-ers style from before. I will try and add some unit tests to it though.

seamus-oconnor: @dainetrinidad I have added tests to the handle*Change methods.

dainetrinidad: Thank you!

PR #1527, opened 2020-01-17
Drilling company migration

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-802


app/backend/wells/migrations/0106_auto_20200116_2327.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 2.2.9 on 2020-01-16 23:27
+
+from django.db import migrations
+
+
+INSERT_DRILLING_COMPANIES_INTO_ORG = """
+    INSERT INTO registries_organization (create_user, create_date, update_user, update_date, org_guid, name, effective_date, expiry_date, province_state_code)
+        SELECT create_user, create_date, update_user, update_date, drilling_company_guid, name, '2018-01-01 00:00:00', '9999-12-31 00:00:00', 'BC' FROM drilling_company;
```
stephenhillier: I don't think all companies are in BC, but I also see that we don't have this info for these old records in our own tables.  I'll ask around and maybe we can get it somewhere?

app/backend/wells/migrations/0106_auto_20200116_2327.py
```diff
@@ -0,0 +1,52 @@
...
+    WHERE org_of_person_responsible_guid IS NULL;
+"""
+
+# Update any legacy submissions to use the same GUID which will point to the drilling companies inserted for the FK-ed well
+UPDATE_LEGACY_SUBMISSION_TO_OLD_DRILLING_COMPANY_GUID = """
+    UPDATE activity_submission s1
+    SET org_of_person_responsible_guid = drilling_company_guid
+    FROM activity_submission s2
+        INNER JOIN well w ON s2.well_tag_number = w.well_tag_number
+    WHERE
```
stephenhillier: Does this just ensure that only activity reports tied to a record in the well table are updated?

app/backend/wells/migrations/0106_auto_20200116_2327.py
seamus-oconnor: Yea they are not (I saw one that has "AB" in the name). When I was chatting with Lindsay — because this is a NOT NULL field — the choices were "BC" or "OTHER". Lindsay suggested that we set them all to "BC" and then later in a different task to manually go through the organizations and clean them up by merging the same ones and updating the province if possible.

app/backend/wells/migrations/0106_auto_20200116_2327.py
seamus-oconnor: Yes. If this SQL isn't run then the legacy records already created will still display (in the UI) nothing for the "company of the person responsible". Maybe I should get you some SQL to run against production to know how many of these legacy records exist, how many have a `org_of_person_responsible` is NOT NULL. Would that be helpful to know before this SQL is run?

PR #1533, opened 2020-01-31
Adds two drill and pump certifications

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-730

image

image


seamus-oconnor:
This PR builds off of #1529 and that PR needs to be merged before this one.

PR #1543, opened 2020-02-15
3D aquifers a.k.a. Vertical Aquifer Extents

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-569

View vertical aquifer extents on well edit screen:
vertical-aquifer-extents-view

Edit a well's vertical aquifer extents:
vertical-aquifer-extents-edit


app/backend/aquifers/models/vertical_aquifer_extents.py
```diff
@@ -0,0 +1,70 @@
...
+        verbose_name='Geo-referenced location of aquifer\'s depth', srid=4326)
+    aquifer = models.ForeignKey(
+        Aquifer, db_column='aquifer_id', on_delete=models.PROTECT, blank=False, null=False,
+        db_comment=('The referenced aquifer at this point.'))
+    start = models.DecimalField(
+        db_column='vertical_aquifer_extent_from', max_digits=7, decimal_places=2, verbose_name='From',
+        null=False, blank=False, validators=[MinValueValidator(Decimal('0.00'))])
+    end = models.DecimalField(
+        db_column='vertical_aquifer_extent_to', max_digits=7, decimal_places=2, verbose_name='To',
+        null=True, blank=True, validators=[MinValueValidator(Decimal('0.01'))])
```
stephenhillier: Is elevation available?  This data will be harder to use if we don't store the elevation from the interpreted data.

seamus-oconnor: Elevation information was in the spreadsheet they initially gave us. They said to ignore that column of information and they only wanted to & from ground level. If elevation is derived (as they said in an email) from ground level elevation and well lithology do we need to store elevation as well?

PR #1546, opened 2020-02-21
Adds command to create legacy well records

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-48

E.g. `python manage.py legacy_records --start=1 --end=20`


app/backend/gwells/models/bulk.py
```diff
@@ -51,8 +51,10 @@ class Meta:
     update_to_aquifer = models.ForeignKey(
         Aquifer, db_column='to_aquifer_id', on_delete=models.PROTECT, blank=False,
         null=False, verbose_name='Aquifer ID Number',
+        related_name='update_to_aquifer',
         db_comment=('The aquifer that this well is to be correlated with'))
     update_from_aquifer = models.ForeignKey(
         Aquifer, db_column='from_aquifer_id', on_delete=models.PROTECT, blank=True,
         null=True, verbose_name='Aquifer ID Number',
+        related_name='update_from_aquifer',
```
stephenhillier: Since this will end up being the name of collections of these objects, the default might make more sense here (`update_from_aquifer_set`), but it's not a big deal.

app/backend/wells/management/commands/legacy_records.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    Run from command line:
+    python manage.py legacy_records
+    """
+
+    def add_arguments(self, parser):
+        # Arguments added for debugging purposes.
+        # e.g. don't cleanup, don't upload: python manage.py export_databc --cleanup=0 --upload=0
+        parser.add_argument('--start', type=int, nargs='?', help='Well to start at', default=1)
+        parser.add_argument('--end', type=int, nargs='?', help='Well to end at', default=50)
```
stephenhillier: This looks good to me, but just thinking about running this command in batches, it might help to have an option to just update the next `n` wells ordered by well tag.

app/backend/wells/management/commands/legacy_records.py
```diff
@@ -0,0 +1,76 @@
...
+
+class Command(BaseCommand):
+    """
+    Run from command line:
+    python manage.py legacy_records
+    """
+
+    def add_arguments(self, parser):
+        # Arguments added for debugging purposes.
+        # e.g. don't cleanup, don't upload: python manage.py export_databc --cleanup=0 --upload=0
```
stephenhillier: leftover comment?

app/frontend/src/submissions/components/SubmissionForm/WellHistory.vue
```diff
@@ -68,11 +65,11 @@
...
           </div>
-          <div class="font-weight-bold mt-3">
-            {{create_user}}
-            Created this well on
-            {{create_date | moment("MMMM Do YYYY [at] LT")}}
-          </div>
+        </div>
+        <div class="font-weight-bold mt-3" v-if="showHistory">
+          {{create_user}}
+          Created this well on
```
stephenhillier:
```suggestion
          created this well on
```

app/backend/wells/management/commands/legacy_records.py
seamus-oconnor: The start and end range is easier to compute to a list of well_tag_numbers without an activity submission. If there are `--start` and `--amount` flags then wouldn't the `--amount` imply the number of "well's without an activity submission"? I don't want someone to assume that --amount=5 will actually process 5 wells when in reality the next 5 wells only has 2 without an activity submission.

app/backend/wells/management/commands/legacy_records.py
seamus-oconnor: Yep! I thought I deleted this already... :thinking:

app/backend/wells/management/commands/legacy_records.py
stephenhillier: True, but I'm just thinking in terms of running a series of jobs "process the next 1000 wells that need submission history" 100 times vs "process well 7001 to 8000" then "process 8001 to 9000", then "process 9001 to 10000" etc.

Also, --amount=5 should still process 5 wells (unless we run out of wells to process), but possibly well tag numbers 1 through 10 not 1 through 5.

seamus-oconnor: I've added a `--next=` argument which (optionally) uses `--start` to process X number of records.

PR #1547, opened 2020-02-25
Fixes DB comments

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-888

1. Fixes model import issues.
2. Updates DB with db_comments defined on all models


app/backend/wells/serializers_v2.py
```diff
@@ -63,20 +63,25 @@ def to_representation(self, instance):
...
         errors = {}
-        if latitude == '' or latitude is None:
-            errors['lat'] = ['This field is required.']
-        if longitude == '' or longitude is None:
-            errors['lng'] = ['This field is required.']
+        if well_tag_number:
+            well = Well.objects.get(well_tag_number=well_tag_number)
+            if well.geom:
+                longitude = well.geom.x
+                latitude = well.geom.y
```
stephenhillier: does this mean we won't be able to change the lat/long by submitting a report (if we are referencing an existing well)?

seamus-oconnor: Hum yea at the moment it does. I should check that lat and lng are not in the payload first before falling back to the well for the geom info.

PR #1550, opened 2020-02-26
Fixes string formatting syntax error in legacy_records command

seamus-oconnor:
Fixes a bug when running `python manage.py legacy_command` where the error string isn't being formatted properly.


seamus-oconnor:
I need to make more changes to this command. A new PR will be coming for those changes including this small bug fix in this one.

PR #1551, opened 2020-02-27
Fixes for legacy_records command

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-48

The legacy_records command no longer updates well after LEGACY submission created

### DIFF
--- a/app/backend/wells/management/commands/legacy_records.py
+++ b/app/backend/wells/management/commands/legacy_records.py
@@ -57,7 +57,7 @@ class Command(BaseCommand):
         num_wells = len(wells)
 
         if num_wells == 0:
-            self.stdout.write(self.style.ERROR(f'No records foundd between well tag number {start} and {end}'))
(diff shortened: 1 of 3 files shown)


app/backend/wells/management/commands/legacy_records.py
```diff
@@ -33,33 +35,59 @@ def add_arguments(self, parser):
...
         num_wells = 0
         if to_do_amount:
             wells = self.find_next_n_wells_without_legacy_records(start, to_do_amount)
         else:
             wells = self.find_wells_without_legacy_records(start, end)
 
         num_wells = len(wells)
+
+        if num_wells == 0:
+            self.stdout.write(self.style.ERROR(f'No records foundd between well tag number {start} and {end}'))
```
stephenhillier:
```suggestion
            self.stdout.write(self.style.ERROR(f'No records found between well tag number {start} and {end}'))
```

PR #1553, opened 2020-03-02
Bulk upload of vertical aquifer extents

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-856

image


app/backend/gwells/views/bulk.py
```diff
@@ -40,21 +62,23 @@ def post(self, request, **kwargs):
...
         update_db = 'commit' in request.GET
 
+        incoming_well_tag_numbers = {wtn for aquifer in aquifers for wtn in aquifer['wellTagNumbers']}
+        incoming_aquifer_ids = {aquifer['aquiferId'] for aquifer in aquifers}
+
+        existing_wells = self.lookup_existing_wells(incoming_well_tag_numbers)
+        existing_aquifers = self.lookup_existing_aquifers(incoming_aquifer_ids)
+
+        if len(self.unknown_well_tag_numbers) > 0 or len(self.unknown_aquifer_ids) > 0:
+            return self.return_errors({})
```
dainetrinidad: Is this supposed to be `self.return_errors()`?

seamus-oconnor: I could remove the empty object - but that function (for `BulkWellAquiferCorrelation` [takes a `changes` dict](https://github.com/bcgov/gwells/pull/1553/files#diff-db78b8a1ffad2b2bba9d89d02969f4a3R142). The code will either have to pass an empty dict or init on inside of `return_errors`. I felt it was safer and more explicit to pass the empty dict. Thoughts?

PR #1560, opened 2020-03-04
Expire drilling companies

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-816


seamus-oconnor:
@stephenhillier I am editing the existing migration to insert all the drilling_company records as expired when they are inserted into the registries_organization table. This won't work for staging (which has already run this migration), but this edit will be run on production.

seamus-oconnor:
@stephenhillier (after talking to Daine) in order to test this updated migration we will need to bring down a PROD snapshot of the DB.

dainetrinidad:
The staging database needs to be updated to match prod before this can be merged in.

PR #1568, opened 2020-03-05
Hides well tag number dropdown for construction reports

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-872

image


stephenhillier:
The construction report serializer could also be modified to remove the well tag option.

seamus-oconnor:
@stephenhillier I am not seeing well_tag_number as a field on the `WellConstructionSubmissionSerializer`

PR #1577, opened 2020-03-13
Aquifer map shows all wells that intersect geom

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-533

image


seamus-oconnor:
> Looks good. Just an observation on the screenshot - the uncorrelated color could be a bit lighter to contrast better with the blue wells.

@stephenhillier I have changed the uncorrelated well colour to white. It stands out a bit better now.

PR #1579, opened 2020-03-13
Adds unique constraint to Well model

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-942

image


seamus-oconnor:
I am going to close this PR as it is the wrong approach. New PR coming!

PR #1583, opened 2020-03-17
Better error message for in-use plate numbers

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-942

image


seamus-oconnor:
This PR accomplishes the same thing that PR #1579 tried to do. But this time with no weird unique constraint on activity submissions. :muscle:

PR #1594, opened 2020-04-07
Upgrades aquifer geom from Polygon to MultiPolygon

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-891

image


app/backend/wells/views_v2.py
```diff
@@ -97,7 +97,7 @@ def get_queryset(self):
                 raise NotFound(f'Unknown aquifer {intersects_aquifer_id}')
 
             # Simplify polygon and expand it by 1km
```
dainetrinidad, line 99: Does this comment still reflect the code?

seamus-oconnor: It does. I changed the value of the simplification to match the amount of simplification that is done when an aquifer geom is saved.

app/backend/aquifers/change_history.py
```diff
@@ -0,0 +1,185 @@
...
+    history = []
+    history += get_aquifer_reversion_history(aquifer)
+    history += get_vertical_aquifer_extents_history(aquifer)
+
+    history.sort(key=get_history_date, reverse=True)
+
+    return history
+
+def get_aquifer_reversion_history(aquifer):
+    # pylint: disable=protected-access
```
dainetrinidad, line 54: I really like the docstrings in gwells` change_history.py describing what the function does and what it returns. Can we add them here too?

I also noticed that it shares a similar codebase, generating a history diff. Is there a reason to name them differently (history vs history_diff) and does it make more sense to drop the "diff" in this context?

seamus-oconnor: Yea I can make that change.

seamus-oconnor:
@stephenhillier and @dainetrinidad I have made some more changes to this PR. There is now an API test for aquifer history. And the MultiPolygons are now only exposed in the V2 API. Please review. Thx!

PR #1617, opened 2020-05-05
test tile cache

stephenhillier:


openshift/pg_tileserv/nginx.conf
```diff
@@ -0,0 +1,150 @@
...
+        # Load configuration files for the default server block.
+        include      /opt/app-root/etc/nginx.default.d/*.conf;
+
+        location /gwells/tiles/ {
+            proxy_pass http://localhost:7800/;
+            proxy_set_header Host $host;
+            proxy_set_header X-Real-IP $remote_addr;
+
+            proxy_cache tile_cache;
+            proxy_cache_valid any 30m;
```
seamus-oconnor, line 118: I think this cache time should be shorter until there is explicit cache invalidation.

PR #1641, opened 2020-05-19
Wally Well View Added

AlexZorkin:


seamus-oconnor:
@AlexZorkin no aquifers in WALLY?

AlexZorkin:
We pull data from the api after click so don't need the info baked into the vector layer.

PR #1653, opened 2020-05-29
Bulk Well and Aquifer screen now strips IDs from file name before upload

seamus-oconnor:
https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-1181?

Also bulk well document upload now warns that files could be overridden.

Screenshot from 2020-05-29 12-57-42

### DIFF
--- a/app/frontend/src/bulk/components/BulkAquiferDocuments.vue
+++ b/app/frontend/src/bulk/components/BulkAquiferDocuments.vue
@@ -176,7 +176,7 @@
                       <template slot="documents" slot-scope="row">
                         <ul>
                           <li v-for="(file, index) in row.item.documents" :key="index">
-                            {{ file.name }}
+                            {{ fileNameWithoutPrefix(file.name) }}
                           </li>
                         </ul>
                       </template>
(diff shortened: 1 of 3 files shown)


app/frontend/src/bulk/components/BulkWellDocuments.vue
```diff
@@ -369,9 +410,11 @@ export default {
       }
     },
     reset () {
+      this.willOverwriteExisting = false
       this.apiError = null
       this.apiValidationErrors = {}
       this.isSaving = false
+      this.unknonwnWellIds = null
```
dainetrinidad, line 417: Would it be worth it to fix this typo?

app/frontend/src/bulk/components/BulkAquiferDocuments.vue
```diff
@@ -299,7 +301,7 @@ export default {
       return Object.keys(this.apiValidationErrors).length > 0
     },
     aquiferDocuments () {
-      const list = []
+      const list = {}
```
dainetrinidad, line 304:
```suggestion
      const docs = {}
```

app/frontend/src/bulk/components/BulkWellDocuments.vue
seamus-oconnor: :man_facepalming: arg - yes I will fix it! :-D

PR #1658, opened 2020-06-02
Wells Subsurface API endpoint

dainetrinidad:


app/backend/wells/serializers_v2.py
```diff
@@ -376,3 +376,16 @@ class WellExportAdminSerializerV2(WellExportSerializerV2):
...
+class WellAquiferSerializerV2(serializers.ModelSerializer):
+    """Serializes aquifer info for a selection of wells"""
+    aquifer_subtype = serializers.ReadOnlyField(source='aquifer.subtype.description')
+
+    class Meta:
+        model = Well
+        fields = (
+            "well_tag_number",
+            "aquifer",
+            "aquifer_subtype"
```
seamus-oconnor, line 390: With only these three params this API is very specific to Wally's needs. Should it be expanded to include all the fields of an aquifer so it is a generic as it is a public API?

app/backend/wells/views_v2.py
```diff
@@ -487,3 +488,33 @@ def list(self, request, *args, **kwargs):
...
+
+            for w in wells:
+                if not w.isnumeric():
+                    raise ValidationError(detail='Invalid well')
+
+            wells = map(int, wells)
+
+            qs = qs.filter(well_tag_number__in=wells)
+
+        return qs
```
seamus-oconnor, line 520: I believe the serializer will make a SQL call for each aquifer in the returned results. Add a `.select_related` so all the aquifer data is fetched once?

stephenhillier:
As another option, you could potentially combine this with the screens (?) endpoint that Wally is using and make a "well cross section" endpoint, that might be useful for public use.

dainetrinidad:
> As another option, you could potentially combine this with the screens (?) endpoint that Wally is using and make a "well subsurface data" endpoint (that has all the data that might be needed for anyone to display cross sections, not just us)? That might be useful for public use.

I like this idea, it might require some refactoring. I will reach out to you and @seamus-oconnor on Teams/RC

PR #1710, opened 2020-11-13
Openshift 4 config changes

stephenhillier:


dainetrinidad (approved):
I having separate files for ocp4. Let's test this for staging

app/backend/gunicorn.ocp4.cfg
```diff
@@ -0,0 +1,52 @@
...
+#       process is still communicating and is not tied to the length
+#       of time required to handle a single request.
+#
+#   keepalive - The number of seconds to wait for the next request
+#       on a Keep-Alive HTTP connection.
+#
+#       A positive integer. Generally set in the 1-5 seconds range.
+#
+
+# workers = 18
```
AlexZorkin, line 46: Should this be commented out? What is the default worker count?

openshift/ocp4/README.md
```diff
@@ -0,0 +1,60 @@
...
+Use `oc -n 26e83e-tools policy add-role-to-group system:image-puller system:serviceaccounts:26e83e-dev` (note namespace at end- repeat for test and prod).
+
+### Secrets / ConfigMaps
+
+The following secrets / configmaps need to be deployed to each environment namespace (dev/test/prod):
+| name | kind |
+| gwells-django-secrets| Secret |
+| gwells-database-secrets | Secret |
+| gwells-e-licensing-secrets| Secret |
+| gwells-minio-secrets | Secret |
```
AlexZorkin, line 55: I think this was changed from gwells-minio-secrets to minio-secrets?

AlexZorkin (commented):
That's a lot of config! I gave it a once over but I think the true test of this is if it all deploys properly and runs so nice work Steve.

app/backend/gunicorn.ocp4.cfg
stephenhillier: I did some digging in the repo and `backend.dc.json` sets this through an environment variable to 4 workers. I don't know what the ideal settings are (especially when using gevent workers)... I do remember we had issues with too many workers in the past.

This is the same config as it is now on GWELLS, except for this file doesn't use `chdir = 'backend'` . The newer container for OCP4 has a different folder structure so the chdir setting was leading to the wrong dir.

openshift/ocp4/README.md
stephenhillier: hmm you're right (`minio-access-parameters` is what the deployment configs refer to). However the minio-access-parameters for each environment will be automatically created by the pipeline, so even though its confusing, this currently works as long as `gwells-minio-secrets` is in place.  (we probably wouldn't want to set it up this way again in the future)

app/backend/gunicorn.ocp4.cfg
stephenhillier: I added a comment to this file to show that # workers is configured in the backend.dc.json file.

PR #1720, opened 2020-12-23
Ensure all licences for wells are imported

stephenhillier:
Not just licences associated with an aquifer

Also:  Excel export now populates the licenced_status_code column from the nightly imported licences.

### DIFF
--- a/app/backend/aquifers/management/commands/import_licences.py
+++ b/app/backend/aquifers/management/commands/import_licences.py
@@ -113,11 +113,8 @@ class Command(BaseCommand):
                 code=row['PURPOSE_USE_CODE'].strip(),
                 description=row['PURPOSE_USE'].strip())
 
-        try:
-            # Check to see if licence already exists in DB
-            licence = WaterRightsLicence.objects.get(wrl_sysid=row['WLS_WRL_SYSID'])
-        except WaterRightsLicence.DoesNotExist:
-            licence = WaterRightsLicence(wrl_sysid=row['WLS_WRL_SYSID'])
+
+        licence = WaterRightsLicence(wrl_sysid=row['WLS_WRL_SYSID'])
 
         licence.licence_number = row['LICENCE_NUMBER'].strip()
         licence.quantity_flag = row['QUANTITY_FLAG'].strip()
--- a/app/backend/aquifers/management/commands/import_licences.py
+++ b/app/backend/aquifers/management/commands/import_licences.py
@@ -95,10 +95,10 @@ class Command(BaseCommand):
 
         logging.info("importing licence #{}".format(row['LICENCE_NUMBER']))
 
+        # Check the Licence is for a valid Aquifer and Well
+        # the if check here allows this function to be called with a specific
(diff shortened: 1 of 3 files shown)


app/backend/wells/management/commands/export.py
```diff
@@ -209,8 +210,18 @@ def generate_files(self, zip_filename, spreadsheet_filename):
...
  left join registries_organization on
  registries_organization.org_guid = well.org_of_person_responsible_guid
+
+-- WELL LICENCES: any well having at least 1 licence entry will be marked as Licensed.
+    left join (select well_tag_number, count(*) as cur_licences from well
+    join well_licences on
+    well.well_tag_number = well_licences.well_id
+    group by well_tag_number) as licence_q
+    on well.well_tag_number = licence_q.well_tag_number
+
```
dainetrinidad, line 220: Maybe this comment can be moved out of this sql string? I like the intention, but it leaves the string vulnerable for modification (i.e. a typo or a change in formatting can invalidate it).

app/backend/wells/management/commands/export.py
```diff
@@ -143,11 +143,12 @@ def generate_files(self, zip_filename, spreadsheet_filename):
         #######
         # WELL
         #######
-        well_sql = ("""select well_tag_number, identification_plate_number,
+        well_sql = ("""select well.well_tag_number, identification_plate_number,
```
dainetrinidad, line 146: What was the reason behind specifying the table for this column in here? (Just curious)
This SQL could also be better if we left out ambiguous columns and specified the tables for all of them, even by using an alias `well w`...then `w.well_tag_number, w. identification_plate_number`. No need to change it all now, but once your SQL gets big enough to join multiple tables, it's more prone to bugs and invalid data.

app/backend/aquifers/management/commands/import_licences.py
```diff
@@ -89,20 +89,24 @@ def process_row(self, row, use_dev_fixtures=False, well=None, aquifer=None):
...
             return
 
         logging.info("importing licence #{}".format(row['LICENCE_NUMBER']))
 
         if not well or not aquifer:
             # Check the Licence is for a valid Aquifer and Well
-            aquifer = Aquifer.objects.get(pk=row['SOURCE_NAME'])
-            well = Well.objects.get(pk=row['WELL_TAG_NUMBER'])
+            # the if check here allows this function to be called with a specific
+            # well or aquifer for dev/test environments.
```
dainetrinidad, line 101: The comment pertains to the "if" condition, which I'm guessing is above? Could we move that above the `if` statement to clean it up and provide less confusion?

dainetrinidad (approved):
@stephenhillier Approved! I don't see any big issues but I do have questions and suggestions to clean up some of the code.

app/backend/wells/management/commands/export.py
stephenhillier: the newly joined table `well_licences` also has a column `well_tag_number`, so I had to specify which table to select well_tag_number from.

PR #1730, opened 2021-01-27
update licence status filter (well search page)

stephenhillier:
Ensure that the well search advanced option for Licensed Status references the table of licences.

* note: also need to deprecate the `licenced_status` column


app/backend/aquifers/fixtures/aquifers.json
```diff
@@ -1358,5 +1358,6 @@
         "end": "144.00",
         "geom":  "POINT(-123.39371 49.36007)"
     }
-  }
+  },
+  {"model": "wells.well_licences", "pk": 1, "fields": {"well": 123, "waterrightslicence": 1}}
```
dainetrinidad, line 1362: Can you run lint formatting for this file please? Thanks!

dainetrinidad (approved):
Looks good, but make sure the API tests are passing before merging it in. 

The only other thing I would suggest to add is an additional test case for filtering on unlicensed wells, in the odd case that the code breaks and always returns a license even if it doesn't have one.

app/backend/aquifers/migrations/0036_auto_20210202_1727.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('aquifers', '0035_aquifer_retire_date_20200522_2157'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='aquifer',
+            name='mapping_year',
+            field=models.PositiveIntegerField(blank=True, help_text='Use the following format: <YYYY>', null=True, validators=[django.core.validators.MinValueValidator(1990), django.core.validators.MaxValueValidator(2021)], verbose_name='Date of Mapping'),
```
dainetrinidad, line 17: Would it be better to write a custom validator that ensures no future dates (in this case, just the current year) can be entered?
https://docs.djangoproject.com/en/3.1/ref/validators/

stephenhillier: I agree...   What do you think of tackling this bug in another issue/PR? I made https://apps.nrs.gov.bc.ca/int/jira/browse/WATER-1621 to track it.

The migration was auto-created by Django; I can leave it in to at least bump the max value from 2020 to 2021 or I can remove it from the PR.

PR #1744, opened 2021-02-18
Dan/water 1589

danbigthink:
migration has instructions on manual review process
backend/wells/migrations/0123_retire_wells_sub_class_codes_water_1589.py


danbigthink:
updated code in branch, removed the update of wells/activity_submission where geotech/special was being changed to cls_lp_geo/na

stephenhillier (approved):

PR #1745, opened 2021-02-18
update ocp4 Jenkinsfile, enable more nightly jobs

stephenhillier:


stephenhillier:
note:  some extra comments as well as whitespace has been removed because Jenkins has a limit to how much code can be in the pipeline.

dainetrinidad (approved):

PR #1751, opened 2021-02-25
Dan/water 1232

danbigthink:


app/frontend/src/submissions/components/SubmissionForm/ActivitySubmissionForm.vue
```diff
@@ -41,6 +41,9 @@ Licensed under the Apache License, Version 2.0 (the "License");
       </b-row>
     </div>
     <div v-else>
+      <b-alert show v-if="isStaffEdit && isUnpublished" variant="info">
+        This well is unpublished and will be hidden from DataBC, iMapBC, GWELLS Well Search, and the CSV/XLS export
```
stephenhillier, line 45:
```suggestion
        This well is unpublished and will be hidden from DataBC, iMapBC, GWELLS Well Search, and the CSV/XLS export.
```

app/frontend/src/wells/views/WellDetail.vue
```diff
@@ -39,7 +39,9 @@ Licensed under the Apache License, Version 2.0 (the "License");
             </a>
           </p>
         </b-alert>
-
+        <b-alert show v-if="show.edit && isUnpublished" variant="warning">
+          This well is unpublished and will be hidden from DataBC, iMapBC, GWELLS Well Search, and the CSV/XLS export
```
stephenhillier, line 43:
```suggestion
          This well is unpublished and will be hidden from DataBC, iMapBC, GWELLS Well Search, and the CSV/XLS export.
```

stephenhillier (commented):
The `well_publication_status` field is inconsistent between the well edit and well summary endpoints. Maybe we should reconsider the `is_published` field (or simply add the `well_publication_status_code` field to give us a consistent code to compare to)?

dainetrinidad:
> The `well_publication_status` field is inconsistent between the well edit and well summary endpoints. Maybe we should reconsider the `is_published` field (or simply add the `well_publication_status_code` field to give us a consistent code to compare to)?

In addition to this, I believe the original development checklist asks to revise the API endpoint to have an `is_published` flag. 

**WATER-1232:**
> V2 of the well details API will need to return a is_published flag. (It currently returns well_publication_status which is the description text made for humans not a useful flag to show / hide a notice)

danbigthink:
Updated backend to return is_published flag from both endpoints, the frontend now uses this flag in both places. This solution is the cleanest!

stephenhillier (approved):

app/frontend/src/wells/views/WellDetail.vue
danbigthink: updated grammar

PR #1752, opened 2021-03-02
needed to call the get_current_year fn rather than pass get_current_y…

danbigthink:
…ear fn reference into model


dainetrinidad (approved):

dainetrinidad:
Looks good, but I wonder if it's possible to add tests? I'd like to see tests for validating a function and a variable/integer, which is what this change seems to be for.

stephenhillier (approved):
Tested, works!

PR #1755, opened 2021-03-08
Dan/water 639

danbigthink:
Tested lots locally.
Ran unit tests in frontend, backend unit tests are still lacking executability.


stephenhillier (approved):
Everything looks good, my one comment is if a user can input artesian pressure in both PSI and ft/head then they could enter 2 conflicting values and it would not be possible to determine which one is correct.  We should ask some of the users and possibly make a small change based on what they say.

app/backend/wells/migrations/0124_create_artesian_conditions_col_update_data_water_639.py
```diff
@@ -0,0 +1,101 @@
...
+UPDATE_ACTIVITY_SUBMISSION_ARTESIAN_CONDITIONS = UPDATE_ARTESIAN_CONDITIONS_BASE.substitute(table_name='activity_submission',
+                                                                                            update_user=WATER_639_USER)
+UPDATE_WELL_ARTESIAN_CONDITIONS = UPDATE_ARTESIAN_CONDITIONS_BASE.substitute(table_name='well',
+                                                                             update_user=WATER_639_USER)
+
+UPDATE_ARTESIAN_FLOW_DUMMY_BASE = Template("""UPDATE $table_name
+    SET artesian_flow = NULL,
+        update_user = '$update_user',
+        update_date = now()
+    WHERE artesian_flow = 0.01;
```
stephenhillier, line 22: These dummy values are being cleaned up, but the actual records with a value of `0.01` should also have `artesian_conditions` set to True based on the first SQL statement, correct (looks like it, just double checking)?

dainetrinidad:
Commenting here that the code duplication that SonarCloud is finding is irrelevant, so we should be OK to merge this in.

dainetrinidad (approved):

danbigthink:
> Everything looks good, my one comment is if a user can input artesian pressure in both PSI and ft/head then they could enter 2 conflicting values and it would not be possible to determine which one is correct. We should ask some of the users and possibly make a small change based on what they say. edit: I will add that as a comment in JIRA and follow up on that.

I've added a comment in jira for this. Good find, the original ticket was missing the need for this, now we ensure we calculate the other field upon load and change (ie. change to psi auto changes and calculates head, vice versa).

stephenhillier (approved):
Looks good.  The API test failure is not related to this PR (config issue on OCP4 for Minio dev environments) so I will get that working.

PR #1757, opened 2021-03-11
created export django command export_welltags_documents_csv. tested l…

danbigthink:
…ocally


daine (approved):
Code looks great, approving it, but it would be nice if we can do these two things before merging:
- Update documentation about the required environment variables
- Change references to `well_tag`/`well_tags` to either `wtn` or `well_tag_number`

Thanks @danbigthink !

PR #1759, opened 2021-03-17
Dan/water 833

danbigthink:


app/backend/gwells/management/commands/export_databc.py
```diff
@@ -45,34 +45,149 @@
 # stable!
 # Casing diameter: For now, grab the smallest diameter (should be using type, but we don't have
 # date right now.)
+# UPDATED FOR WATER-833, explicitly labelled column names (keeping older ones intact)
+#   region was added to the export interface but is purposefully left empty, a new ticket will eventually
+#   contain the added data
```
daine, line 50: Appreciate this comment, thanks!

daine (approved):
Code change looks good to me. This whole file/export process can be better, but that's not part of the scope of change.

PR #1763, opened 2021-03-24
updated Alternative specs submitted in Comments section on Well Summa…

danbigthink:
simple update that fixes display logic for field


daine (approved):
I tried to look back and that's a boolean field, but yeah it's a small change so I'm approving it

PR #1764, opened 2021-03-24
altered column transmissitivity on 3 tables, so now numeric(8,6) so w…

danbigthink:
…e can store decimal place values now. modified the models and tested on frontend, entering the field with decimal place values now works. example: 12.123456 works, meanwhile 123.123123 does NOT, neither does 12.1231231


daine (approved):
LGTM - Sonarcloud is complaining about duplicate lines but I don't see any, so please ignore that

PR #1769, opened 2021-03-26
simple fix, change feet to feet bgl on well related data for field to…

danbigthink:
…tal depth drilled


dainetrinidad (approved):
Thanks for making it consistent (feet vs ft)

PR #1774, opened 2021-03-31
removed old well detail django html template, cleaned up views, filte…

danbigthink:
removed old well detail django html template, cleaned up views, filters, renderers, models, serializers for the removal of the well.surface_seal_length field. Added 3 migrations, 1 for altering the well export views, 1 for updating the data as per the ticket specs and 1 for removal of the actual column


app/backend/aquifers/fixtures/aquifers.json
```diff
@@ -892,6 +892,7 @@
       "street_address": "6647 CHRISTENSEN ROAD",
       "city": "ANAHIM LAKE",
       "legal_lot": "A",
+      "legal_lot": "A",
```
daine, line 895:
```suggestion
```

daine (approved):

PR #1778, opened 2021-04-07
added leave unsaved page confirmation

danbigthink:


daine (approved):
The code looks good, ty! Could you confirm with PO/business area that the usage of a window/brower alert box is sufficient? I noticed that in this page, `this.$noty` to display a styled div is used for alerts.

PR #1779, opened 2021-04-08
fixed bug, was using field artesian_flow instead of artesian_conditio…

danbigthink:
…ns in a few places. Also did a bit of refactoring to follow PEP's such as two spaces after comment, unused import and too many lines between


danbigthink:
@daine , looks like sonarcloud failed, this is due to the model change I made, it thinks I duplicated code while I just changed a field. Please override this

daine (approved):

daine:
> @daine , looks like sonarcloud failed, this is due to the model change I made, it thinks I duplicated code while I just changed a field. Please override this

You're right. I looked at it the code duplication and it's not correct. Let's ignore that.

PR #1781, opened 2021-04-12
Remove DEFAULT_COLUMNS in the store

danbigthink:
this was a tough bug to resolve, somehow it's possible to modify the value of the const DEFAULT_COLUMNS in the store, super weird, this is why I removed the [...DEFAULT_COLUMNS]. tested and seems ok now


daine:
Thank you for fixing minio on local development as well for docker-compose.

daine (approved):
Code change looks good to me

PR #1887, opened 2022-10-24
Additional info logger and error logger output

Christopher-walsh22:
Adding more debugging output to resolve ticket 1812

### DIFF
--- a/app/backend/gwells/documents.py
+++ b/app/backend/gwells/documents.py
@@ -161,7 +161,7 @@ class MinioClient():
 
     def get_documents(self, document_id: int, resource='well', include_private=False):
         """Retrieves a list of available documents for a well or aquifer"""
-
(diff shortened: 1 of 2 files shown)


app/backend/gwells/documents.py
```diff
@@ -161,7 +161,7 @@ def format_object_name(self, object_name: str, document_id: int, resource='well'
 
     def get_documents(self, document_id: int, resource='well', include_private=False):
         """Retrieves a list of available documents for a well or aquifer"""
-
+        logger.info('Get Documents for: document_id:%s, resource: %s, include_private: %s ', document_id, resource, include_private)
```
norrisng-bc, line 164: Not that it really matters, but since GWELLS is on Python 3.6 you could also use f-strings which can be easier to read; e.g. `logger.info(f'Get Documents for: document_id:{document_id}...`

norrisng-bc (approved):

PR #1894, opened 2022-11-22
Document reload

Christopher-walsh22:


dmitri-korin-bcps (approved):
I think the effect of this change is if the first query fails then the second query will always run 5 times because it is called asynchronously. But It does not look like it affects any functionality. I think it is ok for troubleshooting.

app/frontend/src/wells/components/Documents.vue
```diff
@@ -92,16 +92,24 @@ export default {
...
         this.files = response.data
       }).catch((e) => {
         console.error(e)
-        this.error = 'Unable to retrieve file list.'
-        if (retryAttempt < 5) {
-          retryAttempt += 1
-          console.log('Load Files Retry: ' + retryAttempt)
-          this.loadFiles()
+        this.error = 'Error: Refreshing document download'
+        while (this.error && retryAttempt < 5) {
```
norrisng-bc, line 103: *Nitpick:* since `var retryAttempt = 1`, it'll retry 4 times instead of 5

norrisng-bc (approved):

PR #1916, opened 2022-12-21
Feat: CodeWithUs GWELLS Enhancements Dec 2022

AlexZorkin:
**P1T1** - Updated Aquifer Licensing Information Text

**P1T2** - Replaced hydraulically connected text with link to guide document

**P1T3** - Removed hydraulically connected from advanced aquifer search

**P1T4** - Updated aquifer_resource_section code - Updated from 'Artesian advisory' to 'Drilling and operation advisory'

**P1T5** - Added new and updated existing aquifer summary tooltips

**P1T6** - Technical Report field added to Well and Activity Submission - This new field highlights the existence of a technical assessment or Environmental Flow Needs report.

**P1T7** - Well search legend and map updated with new icon - Added a new legend item and map icon for legacy pump test existing on a well.

**P2T1** - Removed pumping stress index - Removed 'Pumping Stress Index' as an option from aquifer search and aquifer summary. I did not remove the aquifer_resource_section_code with matching ‘P’ code because there are aquifer resource reports that use this code as a reference value. Instead we simply remove the option from the frontend options array.

**P2T3** - Added aquifer notations to advanced aquifer search

**P2T2,T4,T5** - Aquifer Notations added to Summary, Search, and Map Views


app/backend/aquifers/migrations/0038_aquifer_resource_section_update.py
```diff
@@ -0,0 +1,19 @@
...
+"""
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('aquifers', '0037_auto_20210224_2343'),
+    ]
+
+    operations = [
+        migrations.RunSQL(UPDATE_AQUIFER_RESOURCE_SECTION)
```
norrisng-bc, line 18: I think we should provide a `reverse_sql` here, just in case we need to rollback the migration.

app/backend/wells/migrations/0137_add_hydraulic_info_postgis_well_view.py
```diff
@@ -0,0 +1,41 @@
...
+    under the postgis_ftw schema in the gwells database. The purpose of this field
+    is to check whether there is legacy pump test data for a well, and then display
+    a different well icon on the well search map if this data exists.
+    """
+    dependencies = [
+        ('wells', '0136_add_technical_report_field'),
+    ]
+
+    operations = [
+        migrations.RunSQL(UPDATE_GWELLS_WELLS_VIEW_SQL)
```
norrisng-bc, line 40: I think we should provide a `reverse_sql` here, just in case we need to rollback the migration.

norrisng-bc (changes requested):

app/backend/aquifers/migrations/0038_aquifer_resource_section_update.py
AlexZorkin: Added reverse migration sql code

app/backend/wells/migrations/0137_add_hydraulic_info_postgis_well_view.py
AlexZorkin: Added reverse migration sql code

norrisng-bc (approved):

PR #1999, opened 2023-08-22
Bug/aquifer id and lithology missing

davidclaveau:


Christopher-walsh22 (commented):
Do we want to remove the excess data being passed into Aquifer data as it is no longer being used in the form?

sanjaytkbabu (approved):

PR #2075, opened 2023-12-06
[GWELLS-2055] FEATURE** Search by licences

LocalNewsTV:
## Pull Request Standards

- [x] Documentation is updated to reflect change

# Description

- Add search parameter to lookup wells with a specific licence
- add new optional table column to Search results to display licence numbers

### DIFF
--- a/app/backend/wells/serializers_v2.py
+++ b/app/backend/wells/serializers_v2.py
@@ -13,7 +13,7 @@
 """
 import logging
 from decimal import Decimal
-
+from django.db import connection
 from rest_framework import serializers
 from django.contrib.gis.geos import Point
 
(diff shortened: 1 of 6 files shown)


davidclaveau (approved):

app/backend/wells/serializers_v2.py
```diff
@@ -286,9 +286,29 @@ class Meta:
             "static_water_level",
             "alternative_specs_submitted",
             "technical_report",
-            "drinking_water_protection_area_ind"
+            "drinking_water_protection_area_ind",
+            "licence_number"
         )
+    def get_licence_number(self, obj):
+        with connection.cursor() as cursor:
+            licence_num_query = """
```
fergmac, line 294: I'm looking for context on why we are using raw SQL here? I understand there are situations where routing around the model layer and using raw SQL is necessary.

PR #2109, opened 2024-01-12
[FEATURE] Code With Us QaQc Dashboard Dec 2023

AlexZorkin:
## Pull Request Standards

# Description

- Added a new 'QaQc Dashboard' to the GWELLS Nav Bar. This is only visible to users with the edit role.
- The QaQc Dashboard view contains 3 new tables accessible via selectable tabs 'Record Compliance', 'Mislocated Wells', and 'Cross Referencing'
- Record Compliance: This table shows any well records that are missing required information.
- Mislocated Wells: This table displays wells that have incongruent location data.
- Cross Referencing: This table displays well records that have been cross referenced by the internal team.

note: This pr is still a work in progress and will be used for testing for the time being.


app/backend/wells/views_v2.py
```diff
@@ -573,3 +577,170 @@ class WellDetail(WellDetailV1):
...
+    ordering = ('well_tag_number',)
+
+    def get_queryset(self):
+        """
+        This view should return a list of all mislocated wells
+        for the currently authenticated user.
+        """
+        queryset = Well.objects.all()
+
+        queryset = Well.objects.select_related('well_status').annotate(
```
fergmac, line 605: For clarification, here we are getting all the Well and related Well Status objects and updating the `work_start_date` and `work_end_date` fields. 

These fields already exist on the Well model, but we are updating them here because legacy wells may not have data for these fields?

app/backend/wells/views_v2.py
```diff
@@ -573,3 +577,170 @@ class WellDetail(WellDetailV1):
...
+            Q(surface_seal_thickness__isnull=True) |
+            Q(aquifer_lithology__isnull=True) |
+            Q(well_status__isnull=True) |
+            Q(work_start_date__isnull=True) |
+            Q(work_end_date__isnull=True) |
+            Q(person_responsible__isnull=True) |
+            Q(company_of_person_responsible__isnull=True) |
+            Q(create_date__isnull=True) |
+            Q(create_user__isnull=True)
+            # Q(natural_resource_region__isnull=True) |
```
fergmac, line 678: Should these lines be removed or uncommented?

app/backend/wells/views_v2.py
```diff
@@ -573,3 +577,170 @@ class WellDetail(WellDetailV1):
...
+    permission_classes = (WellsEditOrReadOnly,)
+    model = Well
+    pagination_class = APILimitOffsetPagination
+
+    # Allow searching on name fields, names of related companies, etc.
+    filter_backends = (WellListFilterBackend, BoundingBoxFilterBackend,
+                       filters.SearchFilter, WellListOrderingFilter, GeometryFilterBackend)
+    ordering = ('well_tag_number',)
+
+    def get_queryset(self):
```
fergmac, line 709: This queryset logic is reused in multiple locations, is it worth pulling this logic out into a common queryset util?

app/backend/wells/utils.py
```diff
@@ -0,0 +1,151 @@
...
+    # Load response into GeoDataFrame
+    data = response.json()
+    regions_gdf = gpd.GeoDataFrame.from_features(data["features"])
+
+    if regions_gdf.empty:
+        print("No natural resource regions found near well location.")
+        return None
+
+    # Assuming the well can only be in one region, return the name of the first region found
+    return regions_gdf.iloc[0]['properties']['REGION_ORG_UNIT_NAME']
```
fergmac, line 86: Here we are checking if `regions_gdf` exists do we also need to check if `iloc` exists before indexing?

app/backend/wells/views_v2.py
AlexZorkin: The Well model doesn't have these two fields available on it so we have to infer the correct start and end dates from the well_status object. The annotation just adds these fields to the queryset object so there's nothing actually being persisted to the database.

app/backend/wells/utils.py
AlexZorkin: the 'regions_gdf.empty' check above is sufficient to prevent indexing errors because it ensures that the dataframe contains data before attempting to access it.

app/backend/wells/views_v2.py
AlexZorkin: fixed, ty

app/backend/wells/models.py
```diff
@@ -1294,6 +1315,50 @@ def update_utm(sender, instance, **kwargs):
         instance.utm_northing = round(utm_point.y)
 
 
+@receiver(pre_save, sender=Well)
```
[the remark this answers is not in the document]
AlexZorkin: good idea, I moved these to signals.py

app/backend/wells/views_v2.py
AlexZorkin: moved to a utils method

app/backend/wells/views_v2.py
fergmac: Ok yes I see that now, I jumped to the wrong class in the file when searching 😄.

fergmac (approved):

PR #2111, opened 2024-01-16
[FEATURE][GWELLS-2094] Address autofill/autocomplete added to Well Owner and Well Location forms

acatchpole:
## Pull Request Standards

# Description

- Added address auto-complete + auto-fill to the Well Owner Information form for well submissions and well editing
- Added address auto-complete + auto-fill to the Well Location form for well submissions and well editing

Utilized the BC Gov Geocoder API to get addresses. This API allows 1000 requests/min so no throttling should occur. BUT, it does not have Postal Code information, so Postal Code is not auto-filled.
 
image

### DIFF
--- a/app/frontend/src/common/constants.js
+++ b/app/frontend/src/common/constants.js
@@ -59,3 +59,5 @@ export const TOOLTIP_TEXT = {
     analysis_method: 'Mathematical solutions (Theis, Cooper-Jacob, Neuman, etc.) where response data (discharge, drawdown/time) from pumping tests are used to estimate the hydraulic properties of aquifers.',
   },
 }
+
+export const GEOCODER_ADDRESS_API = 'https://geocoder.api.gov.bc.ca/addresses.json?q='
--- a/app/frontend/src/common/constants.js
+++ b/app/frontend/src/common/constants.js
@@ -59,5 +59,3 @@ export const TOOLTIP_TEXT = {
     analysis_method: 'Mathematical solutions (Theis, Cooper-Jacob, Neuman, etc.) where response data (discharge, drawdown/time) from pumping tests are used to estimate the hydraulic properties of aquifers.',
   },
 }
-
-export const GEOCODER_ADDRESS_API = 'https://geocoder.api.gov.bc.ca/addresses.json?q='
--- a/app/frontend/src/submissions/components/SubmissionForm/Location.vue
+++ b/app/frontend/src/submissions/components/SubmissionForm/Location.vue
@@ -38,10 +38,23 @@ Licensed under the Apache License, Version 2.0 (the "License");
             id="wellStreetAddress"
             type="text"
             label="Street address"
+            @input="fetchAddressSuggestions"
+            v-on:focus="showList"
+            v-on:blur="hideList"
             :errors="errors['street_address']"
             :loaded="fieldsLoaded['street_address']"
             :disabled="sameAsOwnerAddress"
           ></form-input>
+           <!-- Display the address suggestions -->
+        <div v-if="addressSuggestions.length > 0" class="address-suggestions list-group list-group-flush border" id="address-suggestions-list">
+          <li v-for="(suggestion, index) in addressSuggestions" :key="index">
+            <button @mousedown="selectAddressSuggestion(suggestion)" class="list-group-item list-group-item-action border-0">{{ suggestion }}</button>
+          </li>
+        </div>
+        <!-- Display a loading indicator while fetching suggestions -->
+        <div v-if="isLoadingSuggestions" class="loading-indicator">
+          Loading...
+        </div>
         </b-col>
       </b-row>
       <b-row>
@@ -255,7 +268,9 @@ export default {
   data () {
     return {
       wellAddressHints: [],
-      sameAsOwnerAddress: false
+      sameAsOwnerAddress: false,
+      addressSuggestions: [],
+      isLoadingSuggestions: false
     }
   },
   computed: {
@@ -285,6 +300,73 @@ export default {
       this.streetAddressInput = String(this.ownerMailingAddress)
       this.cityInput = String(this.ownerCity)
     }
+  },
+  methods: {
+    async fetchAddressSuggestions() {
+
+if (!this.streetAddressInput) {
+  this.addressSuggestions = [];
+  return;
+}
+
+this.isLoadingSuggestions = true;
+const params = {
+  minScore: 50,
+  maxResults: 5,
+  echo: 'false',
+  brief: true,
+  autoComplete: true,
+  addressString: this.streetAddressInput
+};
+
+const querystring = require('querystring');
+const searchParams = querystring.stringify(params);
+try {
+  const response = await fetch(`https://geocoder.api.gov.bc.ca/addresses.json?q=${searchParams}`);
+  console.log(encodeURIComponent(params));
+  console.log(searchParams);
+  const data = await response.json();
+  console.log(data);
+  if (data && data.features) {
+    
+    this.addressSuggestions = data.features.map(item => item.properties.fullAddress);
+  } else {
+    this.addressSuggestions = [];
+  }
+} catch (error) {
+  console.error(error);
+  this.addressSuggestions = [];
+} finally {
+  this.isLoadingSuggestions = false;
+}
+},
+selectAddressSuggestion(suggestion) {
+const wellAddressArray = suggestion.split(',');
+
+let province = '';
+
+switch (wellAddressArray.length) {
+  case 3: {
+    this.streetAddressInput = wellAddressArray[0];
+    this.cityInput = wellAddressArray[1];
+    break;
+  }
+  case 2: {
+    this.cityInput = wellAddressArray[0];
+    this.streetAddressInput = '';
+    break;
+  }
+}
+},
+clearAddressSuggestions () {
+this.addressSuggestions = [];
+},
+showList() {
+      document.getElementById('address-suggestions-list').style.display = 'block';
+},
+hideList() {
+document.getElementById('address-suggestions-list').style.display = 'none';
+}
   }
 }
 </script>
--- a/app/frontend/src/submissions/components/SubmissionForm/Location.vue
+++ b/app/frontend/src/submissions/components/SubmissionForm/Location.vue
@@ -220,6 +220,8 @@ import inputBindingsMixin from '@/common/inputBindingsMixin.js'
 
 import BackToTopLink from '@/common/components/BackToTopLink.vue'
 
+import { GEOCODER_ADDRESS_API } from '@/common/constants';
+
 export default {
   name: 'Location',
   mixins: [inputBindingsMixin],
@@ -322,7 +324,7 @@ export default {
       const querystring = require('querystring');
       const searchParams = querystring.stringify(params);
       try {
-        const response = await fetch(`https://geocoder.api.gov.bc.ca/addresses.json?q=${searchParams}`);
+        const response = await fetch(`${GEOCODER_ADDRESS_API}${searchParams}`);
         const data = await response.json();
         if (data && data.features) {    
           this.addressSuggestions = data.features.map(item => item.properties.fullAddress);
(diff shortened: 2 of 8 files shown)


acatchpole:
The code is largely repeated between the two changed components, but there are enough small differences that breaking them out would be onerous.

The API tests that failed on the first run, succeeded when re-run

app/frontend/src/submissions/components/SubmissionForm/Location.vue
```diff
@@ -38,10 +38,23 @@ Licensed under the Apache License, Version 2.0 (the "License");
...
             :loaded="fieldsLoaded['street_address']"
             :disabled="sameAsOwnerAddress"
           ></form-input>
+           <!-- Display the address suggestions -->
+        <div v-if="addressSuggestions.length > 0" class="address-suggestions list-group list-group-flush border" id="address-suggestions-list">
+          <li v-for="(suggestion, index) in addressSuggestions" :key="index">
+            <button @mousedown="selectAddressSuggestion(suggestion)" class="list-group-item list-group-item-action border-0">{{ suggestion }}</button>
+          </li>
+        </div>
+        <!-- Display a loading indicator while fetching suggestions -->
```
LocalNewsTV, line 54: Consider adding some custom CSS to override the default behaviour of the boostrap class so it sits above the rest of the elements instead of pushing them down. Just needs some simple overrides, but don't use a z-index as high as I did :)
```css
specificity-of-elements-to-override-nicely {
  position: absolute;
  z-index: 1000000;
}
```

Example:
image

Current (Spasms as lines are regenerated each request):
image

app/frontend/src/submissions/components/SubmissionForm/Location.vue
```diff
@@ -285,6 +302,97 @@ export default {
...
+  },
+  methods: {
+      /**
+     * Asynchronously fetches address suggestions based on the owner's address input.
+     * If no input is provided, it clears the current suggestions.
+     * On success, it maps the received data to full addresses and updates the addressSuggestions state.
+     * On failure, it logs the error and clears the current suggestions.
+     * Finally, sets the loading state to false.
+     */
+    async fetchAddressSuggestions() {
```
LocalNewsTV, line 314: I use determine a minimum length of characters (3?) before hitting the API with requests. 
You likely won't get useful feedback with a low character count, and its generating more network traffic from the user device to the api

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
```diff
@@ -135,14 +145,108 @@ export default {
...
+    },
+
+    /**
+     * Hides the address suggestions list in the UI.
+     */
+    hideList() {
+      if(document.getElementById('address-suggestions-list')){
+        document.getElementById('address-suggestions-list').style.display = 'none';
+      }
+    }
```
LocalNewsTV, line 243:
```suggestion
    showList(show) {
      if(document.getElementById('address-suggestions-list')){
        document.getElementById('address-suggestions-list').style.display =  show? 'block' : 'none';
      }     
    },
```
Couldn't this be reduced to one function taking a boolean, and reduce code duplication / overall complexity?

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
```diff
@@ -135,14 +145,108 @@ export default {
...
+      };
+
+      const querystring = require('querystring');
+      const searchParams = querystring.stringify(params);
+      try {
+        const response = await fetch(`${GEOCODER_ADDRESS_API}${searchParams}`);
+        const data = await response.json();
+        if (data && data.features) {
+          
+          this.addressSuggestions = data.features.map(item => item.properties.fullAddress);
```
LocalNewsTV, line 186:
```suggestion
        if (data && data.features) {
          this.addressSuggestions = data.features.map(item => item.properties.fullAddress);
```

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
```diff
@@ -135,14 +145,108 @@ export default {
...
+     * Processes the selected address suggestion.
+     * Splits the suggestion into components and updates the owner's province, city, and address inputs accordingly.
+     * Clears the address suggestions afterward.
+     * @param {string} suggestion - The selected address suggestion. ("1234 Street Rd, Name of City, BC")
+     */
+    selectAddressSuggestion(suggestion) {
+      const ownerAddressArray = suggestion.split(',');
+      if(ownerAddressArray){
+        if(ownerAddressArray[ownerAddressArray.length -1].toUpperCase().trim() === 'BC' || ownerAddressArray[-1].toUpperCase().trim() === 'BRITISH COLUMBIA')
+        {
```
LocalNewsTV, line 208:
```suggestion
        if(ownerAddressArray[ownerAddressArray.length -1].toUpperCase().trim() === 'BC' || ownerAddressArray[-1].toUpperCase().trim() === 'BRITISH COLUMBIA') {
```

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
```diff
@@ -135,14 +145,108 @@ export default {
...
+          this.ownerProvinceInput = this.codes.province_codes[0].province_state_code;
+          this.ownerAddressInput = '';
+        }
+      else this.ownerProvinceInput = "";
```
LocalNewsTV, line 212:
```suggestion
      else {
          this.ownerProvinceInput = "";
      }
```

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
```diff
@@ -135,14 +145,108 @@ export default {
...
+      this.ownerCityInput = ownerAddressArray[ownerAddressArray.length -2];
+      if(ownerAddressArray[ownerAddressArray.length -3]) this.ownerAddressInput = ownerAddressArray[ownerAddressArray.length -3];
```
LocalNewsTV, line 214: For readability/maintainability you should declare some variables in the function since you're using -1, -2, -3 to get

```js
const namedValue1 = ownerAddressArray.length - 1;
const namedValue2 = ownerAddressArray.length - 2;
const namedValue3 = ownerAddressArray.length - 3;
```

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
```diff
@@ -135,14 +145,108 @@ export default {
...
+      }
+
```
LocalNewsTV, line 216:
```suggestion
```

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
LocalNewsTV: While technically incorrect, this provides easy opportunity for bugs in future work

LocalNewsTV (changes requested):
Overall Thoughts
- I would use `our` API as a proxy to the geocoder api (new endpoint), and set the geocoder api as an environment variable. Doing so achieves the following:
    - Removes power from the frontend of the app, preserving it in the backend where we control it
    - having it as an `env` variable lets us update the url in the event the architecture changes, without having to do an app rebuild, or open a PR.
    - The api advertises API keys available for production apps like ours, and if that came gets implemented, we would be able to protect that with our API with significantly less overhaul in the future.
    - Can use the existing APIService Class that exists in Vue, and simplify the codebase further by reusing code
    - Good practice 
- fetchAddressSuggestions could be moved into a class or a utilities folder, doing so would reduce complexity in this file, and be more reusable in the application

app/frontend/src/submissions/components/SubmissionForm/Location.vue
```diff
@@ -38,10 +38,23 @@ Licensed under the Apache License, Version 2.0 (the "License");
...
             :errors="errors['street_address']"
             :loaded="fieldsLoaded['street_address']"
             :disabled="sameAsOwnerAddress"
           ></form-input>
+           <!-- Display the address suggestions -->
+        <div v-if="addressSuggestions.length > 0" class="address-suggestions list-group list-group-flush border" id="address-suggestions-list">
+          <li v-for="(suggestion, index) in addressSuggestions" :key="index">
+            <button @mousedown="selectAddressSuggestion(suggestion)" class="list-group-item list-group-item-action border-0">{{ suggestion }}</button>
+          </li>
+        </div>
```
LocalNewsTV, line 53: The `<li>` isn't in a `ul` or `ol` If you're just trying to make rows you can use a table, or divs.

LocalNewsTV (changes requested):

LocalNewsTV (changes requested):

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
acatchpole: implemented

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
acatchpole: fixed

app/frontend/src/submissions/components/SubmissionForm/Location.vue
acatchpole: implemented

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
acatchpole: fixed

app/frontend/src/submissions/components/SubmissionForm/Location.vue
acatchpole: implemented

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
acatchpole: corrected

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
```diff
@@ -135,14 +145,108 @@ export default {
...
+    /**
+     * Clears the current list of address suggestions.
+     */
+    clearAddressSuggestions () {
+      this.addressSuggestions = [];
+    },
+
+    /**
+     * Shows the address suggestions list in the UI.
```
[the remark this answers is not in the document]
acatchpole: corrected

app/frontend/src/submissions/components/SubmissionForm/Location.vue
```diff
@@ -38,10 +38,23 @@ Licensed under the Apache License, Version 2.0 (the "License");
...
             label="Street address"
+            @input="fetchAddressSuggestions"
+            v-on:focus="showList(true)"
+            v-on:blur="showList(false)"
             :errors="errors['street_address']"
             :loaded="fieldsLoaded['street_address']"
             :disabled="sameAsOwnerAddress"
           ></form-input>
+           <!-- Display the address suggestions -->
+        <div v-if="addressSuggestions.length > 0" class="address-suggestions list-group list-group-flush border" id="address-suggestions-list">
```
LocalNewsTV, line 49: you may need to change the name of ID `address-suggestions-list` to something more unique to its use case, its being declared in two separate files. 
Unless I am mistaken, both these pages render on the same page in the application, and the reuse of the same id can pose problems in the DOM

app/frontend/src/common/constants.js
```diff
@@ -59,3 +59,5 @@ export const TOOLTIP_TEXT = {
     analysis_method: 'Mathematical solutions (Theis, Cooper-Jacob, Neuman, etc.) where response data (discharge, drawdown/time) from pumping tests are used to estimate the hydraulic properties of aquifers.',
   },
 }
+
+export const GEOCODER_ADDRESS_API = 'https://geocoder.api.gov.bc.ca/addresses.json?q='
```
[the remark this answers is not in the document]
sanjaytkbabu: removed

sanjaytkbabu (approved):

app/frontend/src/submissions/components/SubmissionForm/Owner.vue
```diff
@@ -127,22 +151,121 @@ export default {
...
+     * On success, it maps the received data to full addresses and updates the addressSuggestions state.
+     * On failure, it logs the error and clears the current suggestions.
+     * Finally, sets the loading state to false.
+     */
+    async fetchAddressSuggestions() {
+      const MIN_QUERY_LENGTH = 3;
+      if (!this.ownerAddressInput || this.ownerAddressInput.length < MIN_QUERY_LENGTH) {
+        this.addressSuggestions = [];
+        return;
+      } else {
```
LocalNewsTV, line 183: If the `if` has a return in it, you don't need the else

LocalNewsTV (changes requested):

LocalNewsTV (approved):

Issue #2125, opened 2024-01-25
Expand Dummy Data in Dev Environment


LolandaE:
**Describe the task**
Import data from PROD to DEV environment. Need to check what can be imported and ensure there is no sensitive data. Alternately expand dummy data that is currently in DEV. 

First option: Point local environment to get data from staging d/b. If this doesn't work, then next option is to create a new database for local that updates alongside prod and staging.

**Purpose**
Small dataset in local and dev environments made some testing hard

**Acceptance Criteria**
- [x] Given developer working in local environment, they have access to a complete dataset

**Additional context**
- Action from Jan 9 retro

davidclaveau:
I was able to create a `TransportServerClaim` on GWELLS' staging namespace that exposes a port for the database for external connections. This setup is mostly used for cluster-to-cluster migrations, but works for our needs. More information on the build can be found here: https://developer.gov.bc.ca/docs/default/component/platform-developer-docs/docs/platform-architecture-reference/set-up-tcp-connectivity-on-private-cloud-openshift-platform/

Follow up: we might need to look into certain restrictions for this setup in the future, including but not limited to:
- cloning the staging database to create a pod for these local connections
- making these local connections "read-only" to ensure data isn't erroneously deleted

PR #2194, opened 2024-03-15
[FEATURE][GWELLS-#2125] - Add the option to run docker-compose with the staging database

davidclaveau:
## Pull Request Standards

# Description

- To allow the user to have more data on a local build, they can now run `docker-compose --env-file ./.env.test up` to point the backend to staging's database
- Updated `.env` for the variables that are swapped when pointing to staging instead
- Create a bash script to differentiate between `local` and `test` to not run certain `command`s depending on the environment.
- Update to README to accommodate the changes


.env
```diff
@@ -1 +1,7 @@
+# .env for local
+ENVIRONMENT=local
 API_TARGET=http://backend:8000/
+DATABASE_USER=gwells
+DATABASE_PASSWORD=test1
+GWELLS_SERVICE_HOST=db
+GWELLS_SERVICE_PORT=5432
```
LocalNewsTV, line 7: We should remove the env from git entirely. We can update the `vue.config.js` file to reflect this. 
To keep an easy template I've historically just saved these as `.env.template` instead so the user can copy paste them and add the correct values without fear of being committed

docker-compose.yml
```diff
@@ -141,22 +144,26 @@ services:
...
       DJANGO_ADMIN_URL: admin
       DJANGO_DEBUG: "True"
       DJANGO_SECRET_KEY: secret
       ENABLE_ADDITIONAL_DOCUMENTS: "True"
       ENABLE_AQUIFERS_SEARCH: "True"
       ENABLE_GOOGLE_ANALYTICS: "False"
-      GWELLS_SERVICE_HOST: db
-      GWELLS_SERVICE_PORT: 5432
+      GWELLS_SERVICE_HOST: "${GWELLS_SERVICE_HOST}"
+      GWELLS_SERVICE_PORT: "${GWELLS_SERVICE_PORT}"
```
LocalNewsTV, line 166:
```suggestion
      GWELLS_SERVICE_HOST: "${GWELLS_SERVICE_HOST:-db}"
      GWELLS_SERVICE_PORT: "${GWELLS_SERVICE_PORT:-5432}"
```
You should be able to add default arguments to your env variables by doing something like this.
Tightens the docker compose, and removes the need for that default .env that risks commits

LocalNewsTV (approved):

PR #2227, opened 2024-05-13
[FIX][GWELLS-2226] fix docker compose is generating an error in local

acatchpole:
## Pull Request Standards

# Description

When running docker-compose up in local, the following error is generated in the backend container : Could not fetch URL https://pypi.python.org/simple/pip/: There was a problem confirming the ssl certificate: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:852) , and the backend does not run.

- One solution is to add trusted host flags when upgrading pip seen [here](https://stackoverflow.com/questions/25981703/pip-install-fails-with-connection-error-ssl-certificate-verify-failed-certi). As manually assigning trusted hosts is not an ideal solution,
a solution was found that did not require their use.

### DIFF
--- a/app/frontend/package-lock.json
+++ b/app/frontend/package-lock.json
@@ -16290,11 +16290,6 @@
       "resolved": "https://registry.npmjs.org/vue/-/vue-2.6.10.tgz",
       "integrity": "sha512-ImThpeNU9HbdZL3utgMCq0oiMzAkt1mcgy3/E6zWC/G6AaQoeuFdsl9nDhTDU3X1R6FK7nsIUuRACVcjI+A2GQ=="
     },
-    "vue-analytics": {
-      "version": "5.16.4",
-      "resolved": "https://registry.npmjs.org/vue-analytics/-/vue-analytics-5.16.4.tgz",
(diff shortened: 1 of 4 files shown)


app/frontend/package-lock.json
```diff
@@ -16290,11 +16290,6 @@
       "resolved": "https://registry.npmjs.org/vue/-/vue-2.6.10.tgz",
       "integrity": "sha512-ImThpeNU9HbdZL3utgMCq0oiMzAkt1mcgy3/E6zWC/G6AaQoeuFdsl9nDhTDU3X1R6FK7nsIUuRACVcjI+A2GQ=="
     },
-    "vue-analytics": {
```
LocalNewsTV, line 16293: Side effects from other tickets?

LocalNewsTV (approved):

app/frontend/package-lock.json
acatchpole: yeah... as this ticket was resolving errors during container building, and this remnant in the package-lock did provoke a warning, i thought it acceptable to include the cleanup in this ticket.

fergmac (approved):

PR #2641, opened 2026-04-24
task: Mostly finished Header migration

raarielgrace:
## Pull Request Standards

- [ ] The title of the PR is accurate
- [ ] The title includes the type of change [`HOTFIX`, `FEATURE`, `etc`]  
- [ ] The PR title includes the ticket number in format of `[GWELLS-###]`
- [ ] Documentation is updated to reflect change [`README`, `functions`, `team documents`]

# Description

- Migrated Header to Primevue and MOSTLY styled it
- All functionality is working! And most of the responsive design stuff is working (so elements appear and disappear as the window is expanded or shrunk)


app/frontend/src/common/components/Header.vue
```diff
@@ -148,4 +173,9 @@ export default {
     height: 0px !important;
   }
 }
+.be-blue {
```
raarielgrace, line 176: This is temporary. We need to change the theme variables later.

app/frontend/src/main.js
```diff
@@ -16,6 +16,15 @@ import { createApp } from "vue";
...
 // import VueNoty from "vuejs-noty";
+import { createPinia } from "pinia";
+import PrimeVue from 'primevue/config';
+import Aura from '@primeuix/themes/aura';
+import { 
+  Button, InputText, Card, Message, Panel, Select, RadioButton,
+  RadioButtonGroup, Checkbox, CheckboxGroup, Listbox, Paginator,
+  Tabs, TabList, Tab, Badge, Popover, Menubar
+} from 'primevue';
+import { Form } from '@primevue/forms';
```
wilwong89: I think for performance reasons the components should be imported on the component level rather than at the app level. Common ones like Button or Card etc are fine here, but something heavier like Forms or Paginator might be better in that particular component.

raarielgrace: Ok, I removed the components you mentioned and everything I didn't think would get rendered on all/most pages. But forms are pretty common, so it might make sense to put Form back here later.

wilwong89: Looks good!

wilwong89 (approved):

